PR #27, opened 2021-04-02
Add pack-mac script

mjacobus:
Hi Charles 👋 

For some reason I could not run the latest release on my mac: 0.8.92

Error on mac

I was able to pack it locally, and then it worked. Running the newly added `npm pack-mac` did the trick.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "pack-win-test": "electron-builder --dir --win",
     "pack-linux-test": "electron-builder --dir --linux",
     "pack-linux": "electron-builder --linux",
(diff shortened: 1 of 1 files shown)


sircharlo:
Hi @mjacobus,
Thanks so much for your contribution!
I'll merge it right away. 
Quick question, did you try running something like [this](https://stackoverflow.com/a/65010167) on the downloaded app?

mjacobus:
Hey @sircharlo 

I did not try that, but that would probably work. I will try that next time. Thank you for sharing the link and for accepting my huge code change :)

PR #156, opened 2022-01-01
Update pt.json

joaochris:
Updated PT-BR


sircharlo:
Thank you very much @joaochris !!

PR #157, opened 2022-01-02
Update pt-pt.json

joaochris:


sircharlo:
Thanks again 😄  @joaochris

PR #178, opened 2022-01-30
Fix freedesktop category (Utilities → Utility)

TiGR:
According to the freedesktop.org specification[1] proper name for utilities category is "Utility".

[1] https://specifications.freedesktop.org/menu-spec/latest/apa.html

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
       ]
     },
     "linux": {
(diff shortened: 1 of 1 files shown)


sircharlo (approved):
Thanks very much for this!

PR #267, opened 2022-04-26
Update pt.json

joaochris:
*Added tooltip when "click to confirm" is triggered
*Added option to disable hardware acceleration
*Added reminder to update media folders after managing media
*Added one explainer for new manage media method


sircharlo:
Ttttttthank you :)

PR #364, opened 2022-05-20
refactor: move code to seperate files for readability and maintainability

mtdvlpr:
Hi (again),

First of all, this is simply a suggestion and I won't be offended if you choose to decline it.

Why?
I've been wanting to contribute to this project for a while, but everytime I sat down to make myself familiar with the codebase I got overwhelmed. It would be much easier if there was some more structure in the code. That's what I tried to do here. 

What?
I created a constants folder where you can put constant variables or simple functions that are often accessed. I also made a helpers folder where you can divide your code into logical parts. I started with the simple ones (obs, log, converters). This greatly increases the readability of the code, because 1 you can easily find the obs code for example and 2 the media-updater becomes smaller and less overwhelming. 

Testing:
I have tested whether this works of course. I tested the PDF conversion and the OBS integration to the extend I could from home (no external camera). However, I am not perfect and this required a lot of tinkering (plus this is the first time editing your code), so be sure to test it yourself as well if you decide to do something with it.

Conclusion:
Feel free to ignore it all together or rearrange it however you want. A lot more can be done, but this is just a start to give you the idea of what I am suggesting. If you think this is useful, I would be more than willing to help you with this big refactor process. I would suggest doing it one part at a time. I could do an initial setup and then you could review it if you would like that.

### DIFF
--- a/constants/endpoints.js
+++ b/constants/endpoints.js
@@ -1,0 +1,18 @@
+const { name, author } = require('./../package.json')
+
+// JW
+const YEARTEXT = (lang) => `https://wol.jw.org/wol/finder?docid=1102022800&wtlocale=${lang}&format=json&snip=yes`
+const JW_API = `https://b.jw-cdn.org/apis/pub-media/GETPUBMEDIALINKS?output=json`
+const JWPUB_API = "https://b.jw-cdn.org/apis/mediator/v1/media-items/"
+
+// GitHub
+const REPO_URL = `https://github.com/${author}/${name}/`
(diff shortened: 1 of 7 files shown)


sircharlo:
Man this is awesome. As you know, this thing has grown to proportions I wasn't expecting.
I'm not a programmer, and it really shows 😆 

This is great, I love the framework that it lays out. Modular, easy to find, use and reuse code as needed.
I honestly think that every part of the code could be reviewed, simplified and improved. 
I'll take all the suggestions and help that I can get!

I'll start testing your PR right away.

sircharlo:
Considering refactoring, definitely needed. And definitely needs to be done one bit at a time.

PR #396, opened 2022-06-22
chore: create cong, date and webdav modules

mtdvlpr:
This is a big refactor, so extensive testing is needed before it can be moved to production.


sircharlo:
only issue I see is a sort of circular dependency between cong and prefs...

mtdvlpr:
Alright, that should be fixed. I'll look into it.

sircharlo:
Latest commit "fixes" it but only on the surface of course. I just moved the 2 offending functions back to prefs
The circular dependency happens because of validateConfig, I think.

PR #429, opened 2022-08-05
Update es.json

wifilogo:
Added "paragraph" to list of localized strings, for future use.


wifilogo:
Added new Spanish strings and other minor changes.

PR #464, opened 2022-09-14
Updated ICON and English Wording

advenimus:
Rounded edges of icon since this is more fluid with Mac OS, and updated the english json file to stated that you require obs compatibility socket (not the regular version).


sircharlo:
Thanks for the contribution!
I think I'd leave the ICO file untouched, as I'm aiming for consistency with other well-known apps on Windows. 

image


However, the Mac build uses the PNG file if I'm not mistaken, so that icon could definitely be rounded to look more like other native Mac icons.

advenimus:
> Thanks for the contribution! I think I'd leave the ICO file untouched, as I'm aiming for consistency with other well-known apps on Windows.
> 
> image
> 
> However, the Mac build uses the PNG file if I'm not mistaken, so that icon could definitely be rounded to look more like other native Mac icons.

Fair enough, this makes sense as yes Windows still uses square icons 🤦

mtdvlpr:
Hi @advenimus, in v22.9.0-alpha we're using a icns file for the macOS icon. If you want  I'd appreciate it if you can update that one to have rounded corners! You can find it [here](https://github.com/mtdvlpr/meeting-media-manager/tree/refactor/next/build/icons)

advenimus:
I updated from that fork and created a pull request. Thanks!

PR #487, opened 2022-10-24
fix: search

mtdvlpr:
Search was not working correctly. Now it is :)


mtdvlpr:
Because of the protected branch rules you set, I require your approval before I can merge my changes :). You can turn off require approval if you feel like it's more of a hassle than a protection. Up to you.

mtdvlpr:
You, as admin, have the option to force merge, without approval and passed checks

sircharlo (approved):

sircharlo:
Can I turn off require approval just for you?

mtdvlpr:
> Can I turn off require approval just for you?

GitHub says:
> The above settings will apply to administrators and custom roles with the "bypass branch protections" permission. 

So apparently you could give me a role with the "bypass branch protections" permission.

sircharlo:
> You either turn it off or on for all collaborators. But since I'm the only collaborator (as of now), turning it off would only affect me.

Good point. I'll turn it off for now.

PR #606, opened 2022-11-04
use bundled wasm file

mtdvlpr:


mtdvlpr:
@sircharlo, I don't know what went wrong last time, but it seems to work now, so might as well use it again :)

mtdvlpr:
Hey @sircharlo, I managed to get the tests working for MacOS and Windows too :). Now the protected branch rules just need to be updated to reflect that. Instead of requiring test, you should require: test (macos-latest), test (ubuntu-latest) and test (windows-latest).

Small problem. The fetch won't work in Windows tests, so I skip those for Windows.

mtdvlpr:
@sircharlo, I think after this PR it is time for 22.11.0. What do you think?

sircharlo:
> Hey @sircharlo, I managed to get the tests working for MacOS and Windows too :). Now the protected branch rules just need to be updated to reflect that. Instead of requiring test, you should require: test (macos-latest), test (ubuntu-latest) and test (windows-latest).
> 
> Small problem. The fetch won't work in Windows tests, so I skip those for Windows, but at least all other tests are now tested for every platform!
> 
> Weird problem. For some reason macOS needs to wait a lot longer before the shuffle music is stopped. Fixed it by adding an extra delay for macOS. But I am curious whether this is also the case for an actual macOS user.

Awesome! I'll try to do that tomorrow.

sircharlo:
> @sircharlo, I think after this PR it is time for 22.11.0. What do you think?

I think so! Although, I noticed one regression. Notice the white flash when removing the picture:


https://user-images.githubusercontent.com/1275666/200133688-f7bfdefe-ebbc-4952-8384-6929b4bb0e10.mp4

It seems the graceful fade-to-black is gone for pictures and videos.
Any idea why?

mtdvlpr:
Yes, I am reloading the window, because otherwise it doesn't go back to being draggable. It's not a nice solution and it could probably use a check to not reload when in fullscreen. Or if you prefer the graceful fade out over being draggable in between presenting it can be removed.

mtdvlpr:
See https://github.com/sircharlo/meeting-media-manager/blob/b73cc89f68e187feb1c8806bbc8da67307ac8b77/src/renderer/pages/media.vue#L92

sircharlo:
> Yes, I am reloading the window, because otherwise it doesn't go back to being draggable. It's not a nice solution and it could probably use a check to not reload when in fullscreen. Or if you prefer the graceful fade out over being draggable in between presenting it can be removed.

Hmm. Would there be a way instead of:

- zooming by scrolling on the preview image (as is currently the case)
- having the preview image reflect the same thing that is seen in the media window (zoom preview along with actual image)
- pan the actual image by mouse panning the preview image

That way, no mouse manipulation would take place on the media window itself, and the drag even could be left alone.
TBH I hate seeing the couse cursor on the main screen during meetings, and anything that can be done offscreen is (IMO) better.
What do you think?

mtdvlpr:
Definitely agree! I haven't thought about zooming the preview, but that should definitely be possible! Making the panning work externally I haven't figured out yet. It's possible, but the trick is how to implement it cleanly. A couple of possible solutions:
- up/down/left/right (arrow) buttons on top of the preview box
- dedicated keyboard shortcuts (arrows)
- a dialog/popup with control buttons to be opened on button click.

Which has your preference?

mtdvlpr:
Definitely agree! I haven't thought about zooming the preview, but that should definitely be possible! Making the panning work externally I haven't figured out yet. It's possible, but the trick is how to implement it cleanly. A couple of possible solutions:
- up/down/left/right (arrow) buttons on top of the preview box
- dedicated keyboard shortcuts (arrows)
- a dialog/popup with control buttons to be opened on button click.

Which has your preference?

sircharlo:
> Hey @sircharlo, I managed to get the tests working for MacOS and Windows too :). Now the protected branch rules just need to be updated to reflect that. Instead of requiring test, you should require: test (macos-latest), test (ubuntu-latest) and test (windows-latest).
> 
> Small problem. The fetch won't work in Windows tests, so I skip those for Windows, but at least all other tests are now tested for every platform!
> 
> Weird problem. For some reason macOS needs to wait a lot longer before the shuffle music is stopped. Fixed it by adding an extra delay for macOS. But I am curious whether this is also the case for an actual macOS user.

I've updated the protected branch rules! :)

PR #610, opened 2022-11-06
feat: Allow panning image through preview

mtdvlpr:


mtdvlpr:
@sirchrarlo, great news! I think I figured it out. You can now zoom and pin through the preview and the preview will show the same as the media window.

mtdvlpr:
With this, we can indeed remove it from the media window itself so that it is always draggable again.

mtdvlpr:
One thing I am now wondering, would it be beneficial to have a reset button to reset the zoom and pan with a click on the button. What do you think, @sircharlo?

Chipperchap:
Of course I do not know how this feature looks in production stage, but I am really looking forward to it. 
As for the reset button, I wonder if all the screen 'movement' being seen by the audience could be 'complicated' - so once finished looking at the zoomed in portion, instead of the audience seeing the zooming out it just instantly becomes regular size again. I am trying to remember what SoundBox used to do. I think using the preview you would zoom in on the portion desired, then click a 'set' button (or similar) and the audience would only see the modified picture. The same in reverse, the preview was reset and then a 'display' or 'set' button was clicked and the image was back to how it is in the workbook / WT again.... but my memory might be incorrect there. 
(Just my 2 cents)

mtdvlpr:
> Of course I do not know how this feature looks in production stage, but I am really looking forward to it. As for the reset button, I wonder if all the screen 'movement' being seen by the audience could be 'complicated' - so once finished looking at the zoomed in portion, instead of the audience seeing the zooming out it just instantly becomes regular size again. I am trying to remember what SoundBox used to do. I think using the preview you would zoom in on the portion desired, then click a 'set' button (or similar) and the audience would only see the modified picture. The same in reverse, the preview was reset and then a 'display' or 'set' button was clicked and the image was back to how it is in the workbook / WT again.... but my memory might be incorrect there. (Just my 2 cents)

@Chipperchap, I like the idea! I will see what I can do with it. If you want, here's a [test version](https://drive.google.com/file/d/1vMuv1zH4kGj-INVbxpXhDUHC8rCX6kRY/view?usp=share_link) for you to see how it is currently implemented.

sircharlo:
> @sircharlo, great news! I think I figured it out. You can now zoom and pan through the preview and the preview will show the same as the media window.

Wow! Do you ever sleep? 😆  Awesome job

sircharlo:
> > Of course I do not know how this feature looks in production stage, but I am really looking forward to it. As for the reset button, I wonder if all the screen 'movement' being seen by the audience could be 'complicated' - so once finished looking at the zoomed in portion, instead of the audience seeing the zooming out it just instantly becomes regular size again. I am trying to remember what SoundBox used to do. I think using the preview you would zoom in on the portion desired, then click a 'set' button (or similar) and the audience would only see the modified picture. The same in reverse, the preview was reset and then a 'display' or 'set' button was clicked and the image was back to how it is in the workbook / WT again.... but my memory might be incorrect there. (Just my 2 cents)
> 
> @Chipperchap, I like the idea! I will see what I can do with it. If you want, here's a [test version](https://drive.google.com/file/d/1vMuv1zH4kGj-INVbxpXhDUHC8rCX6kRY/view?usp=share_link) for you to see how it is currently implemented. Perhaps after testing it you find it so easy/fast that the set/display function is not needed.

Two minor issues I noticed: 
- sometimes, the preview image and the actual image don't match (see 0:03 and 0:22 in video)
- after resizing the media window, the panning would need to be "refreshed" (0:11)

Other than that, I **love** the zoom/pan feature. It's clean, easy and simple. It's out of the way for those who aren't looking for it and probably wouldn't need/use it, while being right there for those who want it.

This feature has been on my radar since the beginning. I'm so happy you made it a reality! @mtdvlpr 

https://user-images.githubusercontent.com/1275666/200180062-6c3b3ab8-d4c0-4a21-9df1-9432fb5629fc.mp4

PR #616, opened 2022-11-06
add malagasy to langs

mtdvlpr:


sircharlo:
`Module not found: Error: Can't resolve 'dayjs/locale/mg' in '/home/runner/work/***/***/src/renderer/.nuxt'`

Seems like Malagasy isn't part of dayjs' list of locales...

mtdvlpr:
> `Module not found: Error: Can't resolve 'dayjs/locale/mg' in '/home/runner/work/***/***/src/renderer/.nuxt'`
> 
> Seems like Malagasy isn't part of dayjs' list of locales...

That's unfortunate.

mtdvlpr:
https://github.com/sircharlo/meeting-media-manager/discussions/508#discussioncomment-4069599

sircharlo:
Oh! Didn't see you were working on that. I had started by filtering out the languages that don't have `dayjs `support from `LOCAL_LANGS `into a new const, but then I noticed that `LOCAL_LANGS `is actually only used for `dayjs`-related things. Can't we just comment out `mg` from the `LOCAL_LANGS` array?

mtdvlpr:
> Oh! Didn't see you were working on that. I had started by filtering out the languages that don't have `dayjs` support from `LOCAL_LANGS` into a new const, but then I noticed that `LOCAL_LANGS` is actually only used for `dayjs`-related things. Can't we just comment out `mg` from the `LOCAL_LANGS` array?

That's indeed a better idea. Just comment it out until it's supported.

PR #634, opened 2022-11-06
potential fix for paralell downloading

mtdvlpr:


sircharlo:
image

Possible regression: error after clear cache and click blue button

mtdvlpr:
Fixed in #635

PR #648, opened 2022-11-07
feat: allow zoom with double click

mtdvlpr:


sircharlo (changes requested):
Pretty cool! 3 notes:
- The double-click event is active for every preview image, even when not being played.
- Would be cool to add a mouse cursor on hover of preview to show that it is actionable...
- The double click zoom isn't consistent; expected behavior: cycles between 1,2,3,4x. Actual: Kind of random zooms.

mtdvlpr:
The weird thing is, the scale is actually changed correctly (1x, 2x, 3x, 4x, 1x)

mtdvlpr:
The first time, it goes from 1x to 2x, so the zoom is doubled (big change), then from 2x to 3x is a smaller change and 3x to 4x is even smaller. But indeed, the second go around it seems like what should be 3x is actually 1.5x for some reason.

sircharlo:
> The first time, it goes from 1x to 2x, so the zoom is doubled (big change), then from 2x to 3x is a smaller change and 3x to 4x is even smaller. But indeed, the second go around it seems like what should be 3x is actually 1.5x for some reason.

I won't lose sleep over it. 😄

mtdvlpr:
It seems that resetting the zoom by using zoom(1) fixes it. I suspect there's some internal state in the package that only gets updated by using the zoom function

sircharlo:
Awesome! It works perfectly!

sircharlo (approved):

PR #650, opened 2022-11-07
fix: no internet warning at startup

mtdvlpr:


mtdvlpr:
To check if this is valid: Remove the langs.json file and startup the app with only one congregation.

sircharlo (changes requested):
image

Doesn't seem to work, ie. the langs are only being fetched when I open settings, but the language tile is missing it's information until then.

sircharlo:
The internet message is gone though, of course

mtdvlpr:
Ah, I think I found the issue. Before the call to get jw langs is made I check if the user is online, but the online store prop hasn't been initialized yet, so it gives an offline warning. Now I check if online only if an error occurs.

sircharlo:
Still a problem for me, but I think I see why. 

The error is being triggered by the call [here](https://github.com/sircharlo/meeting-media-manager/blob/488cb51ac5deb6edbb58503e1ca86c8c3c76e527/src/renderer/layouts/default.vue#L120) because [`langPath`](https://github.com/sircharlo/meeting-media-manager/blob/488cb51ac5deb6edbb58503e1ca86c8c3c76e527/src/renderer/plugins/jw.ts#L58)is equal to `langs.json`, instead of the full path, which is the result of `join`ing `$appPath()` (equals `.`) and the string `langs.json` [here](https://github.com/sircharlo/meeting-media-manager/blob/488cb51ac5deb6edbb58503e1ca86c8c3c76e527/src/renderer/plugins/jw.ts#L23).

The call triggered [here](https://github.com/sircharlo/meeting-media-manager/blob/488cb51ac5deb6edbb58503e1ca86c8c3c76e527/src/renderer/layouts/default.vue#L342) works without issue.

mtdvlpr:
Oh, yes. The appath is calculated based on the prefs store which isn't initialized yet at mounted

mtdvlpr:
That should be it. Meant to remove the online check earlier, but something went wrong there apparently.

sircharlo (approved):
Yessir! It worked. No more offline error, and `langs.json` gets fetched as needed.

PR #671, opened 2022-11-11
Feat/detect pub availability

mtdvlpr:
Test langs:
- Dutch Sign Language has nothing available
- Abui has only the midweek meeting available

Once it's been established that nothing is available and it is not your selected language, it will be filtered out of the langs list.
Every time you (re)enter the home or settings page, 10 languages will be checked and when you select a new language, it will be checked.


mtdvlpr:
What can be considered is to remove the auto check 10 feature. Firstly, because it saves internet/speed.Secondly, how often do people switch their media lang? Once or twice I suppose. Plus, there's a search function, so it's not like filtering the list really helps them navigate it faster.

sircharlo:
> Once it's been established that nothing is available and it is not your selected language, it will be filtered out of the langs list. 

Awesome! Really practical, I like it.

> Every time you (re)enter the home or settings page, 10 languages will be checked and when you select a new language, it will be checked. 

The more I think about this, the less I see the relevance.

> What can be considered is to remove the auto check 10 feature. Firstly, because it saves internet/speed. Secondly, how often do people switch their media lang? Once or twice I suppose. Plus, there's a search function, so it's not like filtering the list really helps them navigate it faster.

Agreed.

Do you think the yeartext endpoint should also be added to the list of checks? I'm thinking of a scenario where the mwb and w are not available, but the yeartext is. However, I'm not sure such a scenario would actually exist in reality.

Will the periodic langs update (every 3 months) simply replace whatever mwb and w check were done? I think that's what's best, in case a language has had some updates in the 3 month period.

mtdvlpr:
Yes, the call that's done every 3 months overrides the checks that have been made.

mtdvlpr:
Removed the regular check :)

sircharlo:
Interesting issue with extracted multimedia in the case of fallback languages. I think in the big refactor, a big of language logic was lost. Look at `media.ts`, [line 67](https://github.com/sircharlo/meeting-media-manager/blob/38eb856808f5ad07fc7b46ac18c037327d60eab4/src/renderer/plugins/media.ts#L67). Extract.Lang is supposed to be used to get extracted media in a fallback language, if such a language is defined in the JWPUB file. The subsequent `getDbFromJWPUB` call should take into account that the db should be fetched in the fallback language, rather than the default one. You can see that this doesn't work as intended for Abui for example; only one image is extracted from the JWPUB file itself, and all other media (from IN language publications) is not fetched.

mtdvlpr:
I believe it's fixed now

sircharlo:
> I believe it's fixed now

You're so fast it's crazy 😆 

So now the publications get downloaded in the fallback language, true, but the media from those publications still get downloaded in the main language, instead of the fallback language. Example, fetch the ABU MW meeting. The IN lff book gets fetched, but then instead of getting the media from the lff book in IN, it tries in ABU (because it's the main congregation language), which fails.

The media for a publication would need to be downloaded in whatever language the publication itself was downloaded (whether fallback or main).

---

Also, I don't know if this is the desired behavior, but when a publication is downloaded in the fallback language, it's saved in the main language's folder, instead of in the fallback language's folder.

image

mtdvlpr:
Alright, it was quite the puzzle, but I managed to get the item.Link property (like we do for extracts) for multiMedia items too, so now the fallback songs are also correctly fetched. The only failure I now see for Abui is a video that is really unavailable as of now (I checked in the JW Library app).

PR #672, opened 2022-11-11
Feat/website in m3

mtdvlpr:
This is a first setup / proof of concept, to show that it's possible. It still needs to be worked out fully.


mtdvlpr:
For now, there's a jw.org button in the presentaion mode footer that opens a browser window on top of the media window and another controller window on top of the main window. I implemented the scrolling and clicking on things. However some clicks don't translate correctly. Trying to click svgs or icon buttons gives issues especially.

sircharlo:
Interesting proof of concept!
2 questions: 
1. Why are you using a new window, instead of repurposing the media window?
2. Why are you using the `disable-site-isolation-trials` switch? As I understand, this is not something you want in production.

mtdvlpr:
1. Actually didn't think of that. Would probably be better!
2. Is necessary to allow listening for clicks inside an iFrame.

mtdvlpr:
There, much better :). As for the actual implementation. I think the way it is implemented now (a button in the present mode) is actually a good start (might have to rethink the actual location/design though). Since 95% of the usage will be to open jw.org. Actually implementing custom websites from loading, will be a lot harder. 1, because the cleanest way would probably to implement it similarly as a media file (can be played, stopped, renamed, reordered, etc.), which will be complex. 2, because other websites might have other problems with cookies, clicks, etc which can't be all accounted for. 3, allowing the use of custom websites during meetings might be assisting in going beyond the guidelines set by the brothers.

sircharlo:
> There, much better :).

Very nice indeed! I like how clean it is now.

> As for the actual implementation. I think the way it is implemented now (a button in the present mode) is actually a good start (might have to rethink the actual location/design though). 

Agreed. Perhaps the top row of buttons (refresh media items, open media folder, view prefix, manual sort) could be moved to a hamburger menu instead, and the JW.org item added there? Just a thought.

> Since 95% of the usage will be to open jw.org. Actually implementing custom websites, will be a lot harder. 1, because the cleanest way would probably be to implement it similarly as a media file (can be played, stopped, renamed, reordered, etc.), which will be complex. 2, because other websites might have other problems with cookies, clicks, etc which can't be all accounted for. 3, allowing the use of custom websites during meetings might be assisting in going beyond the guidelines set by the brothers. What do you think about this, @sircharlo?

Oh definitely. JW.org will cover most (if not all) use cases during congregation meetings, which is the main focus of M3. If a user wants to show a different site (ie for metropolitan witnessing, etc), it could be done manually on the computer, just not though M3.

sircharlo:
<kbd>Ctrl</kbd> <kbd>-</kbd> works to zoom out, and <kbd>Ctrl</kbd> <kbd>0</kbd> te reset the zoom level, but <kbd>Ctrl</kbd><kbd>+</kbd> doesn't work for some reason. Odd.

mtdvlpr:
> Ctrl - works to zoom out, and Ctrl 0 te reset the zoom level, but Ctrl+ doesn't work for some reason. Odd.

So, that turns out to be an [Electron bug(https://github.com/electron/electron/issues/29996)]. Zooming in can be done with <kbd>Ctrl</kbd> <kbd>Shift</kbd> <kbd>+</kbd>

mtdvlpr:
@sircharlo, what do you think of the position of the jw.org button now? I moved the three icons (refresh, open folder, show prexix) to a menu per your request.

sircharlo:
> @sircharlo, what do you think of the position of the jw.org button now? I moved the three icons (refresh, open folder, show prexix) to a menu per your request.

I like the menu; I think it might be a good idea to put the JW.org button in the menu. My logic is that what is in the "misc" menu are commands that are rarely used, in order to save space and keep things relatively uncluttered. The JW.org button won't be used at every meeting, far from it, so I'm not a fan of having it in my face all the time. 

What are your thoughts on that?

mtdvlpr:
> > @sircharlo, what do you think of the position of the jw.org button now? I moved the three icons (refresh, open folder, show prexix) to a menu per your request.
> 
> I like the menu; I think it might be a good idea to put the JW.org button in the menu. My logic is that what is in the "misc" menu are commands that are rarely used, in order to save space and keep things relatively uncluttered. The JW.org button won't be used at every meeting, far from it, so I'm not a fan of having it in my face all the time.
> 
> What are your thoughts on that?

Makes sense!

sircharlo:
I have a nitpick. 

Could the website controller be locked to be the same dimensions as the media window? To make it so that what you see in the controller window is also what you see in the media window?

mtdvlpr:
> 



> I have a nitpick.
> 
> Could the website controller be locked to be the same dimensions as the media window? To make it so that what you see in the controller window is also what you see in the media window?

Well, I can see a situation where you don't want the controller to be fullscreen even though the media window is (when you want to hide the media window using the button in the presentation mode). That's why I made the scrolling scroll by percentage. Also when you're using 1 screen, you want everything to fit.

mtdvlpr:
@sircharlo, any remaining feedback on this?

mtdvlpr:
Jehovahs cutting the days short (Mt 24:22). I feel it too :)

mtdvlpr:
@sircharlo, I have added a refactor in here. The Electron files are now also TypeScript instead of JavaScript, which means autocomplete and more secure code! This requires an extra step before you can run locally. Therefore I simplified the scripts to use. There's a separate prebuild script that tries to execute the pre-build.sh (which fails on Windows). A compile script to convert the Electron TS files to JS which is stored in src/main/dist. And the regular dev and build scripts. So the prebuild is needed once. And the compile script is needed everytime you change the src/main files.

mtdvlpr:
> Seems great so far; I still want to test it on an external monitor, but haven't had the time yet. (Anyone know where all the hours go? I never seem to have enough in a day.)

If we're gonna release 22.12.0 soon, I'd like this to be included. Had any time to test this yet?

sircharlo:
Does this still work? I'm on the `feat/website-in-m3` branch, I ran `yarn install` and `yarn dev`, but clicking on the JW.org button doesn't do anything anymore for me, whether a 2nd screen is present or not.

mtdvlpr:
See https://github.com/sircharlo/meeting-media-manager/pull/672#issuecomment-1319627343

sircharlo:
Lol, yup.

> You probably missed the new `yarn compile` step. See [#672 (comment)](https://github.com/sircharlo/meeting-media-manager/pull/672#issuecomment-1319627343).

sircharlo:
So, it works with multiple screens the same as it does in windowed mode. 
I would however maybe slap a "BETA" badge on the JW.org menu item.
While the feature itself is functional, I see a few ways that something could go wrong in production. 
For example:

- The navigation gets de-syncronized or doesn't work sometimes (try clicking on the language button, or on "Log In")
- The preview window isn't really a "mirror" or the web view, but a "clone". This to me is problematic, since strictly speaking, the web view might at times show text or layout items slightly differently from the preview window. 

I'd personally prefer if the preview window was scaled and zoomed on load, in the style of: 
1. Get web view dimensions (eg., `1920x1080`)
2. Get preview window dimensions (eg., `1280x720`)
3. Calculate ratio between the two (eg, `2:3` or `0.666666667`)
4. Set the zoom or scale factor of preview window to `ratio` (eg., `0.666666667%`)

That way, the preview window would actually be an accurate reflection of what is displayed on the TVs (albeit scaled, but ). 
I realize this is asking for alot, though. I'm just spitballing here.

sircharlo:
I also think some might find it more useful if the web view had a sort of visual indication of where the mouse pointer is. Sometimes, the point of these demos is to say "click here, click there", and that might be harder for some to understand if they don't see where the pointer is.

mtdvlpr:
All great points! True, there are several points of improvement, one more complex than the other, but it would already be a nice feature as-is, so I like the idea of releasing it now under a BETA disclaimer and continuing tinkering on it.

Issue #744, opened 2022-11-24
[BUG] Missing "The Whole House of Ahab Will Perish" image


mtdvlpr:
**Describe the bug**
For this weeks midweek meeting (21-27 November), the Treasures from God's Word referenced an older Treasures from God's Word which contains a chart called "The Whole House of Ahab Will Perish". This image is not listed as media for the midweek meeting.

**To Reproduce**

1. Go to 'Home'
2. Click on 'Fetch media'
3. Do 'open presentation mode'
4. See that the chart is missing

**Expected behavior**
I would expect the referenced chart to be included in the media (it is in JW Library)

**Screenshots**
<details><summary>See image</summary>

![image](https://user-images.githubusercontent.com/46671786/203737364-f7ad13cb-ec52-4d66-85d5-55dd4ed05d6b.png)

</details>

**System specs (please complete the following information):**

- OS: Any
- App Version 22.11.0

**Additional context**
@sircharlo, since you have a deeper knowledge of the fetching mechanism, maybe you know where this image could be extracted if it is at all possible.

PR #745, opened 2022-11-24
fix: process refs of internal refs

mtdvlpr:


mtdvlpr:
@sircharlo, I think I managed to fix #744 here. Here I also process the internal references of the internal references. So the mwb jwpub file references the treasures part and the treasures part references another treasures part from another place. It will now also include the images of that sub reference. My only question is, does this always do what we want or will it sometimes get too much media? What do you think?

sircharlo:
Hey, sorry I've been out of it. I'm slowly recovering from COVID. I'll try to test this branch against the next few weeks of media and see what the cat drags in, if it's too generous of a fetch. I seem to remember not wanting this behavior when I started investigating the fetch logic a few years back, but I can't remember why exactly.

mtdvlpr:
Sorry to hear! Hope you're feeling better. Take your time. I will also test it further (although internal refs inside internal refs don't happen that often).

mtdvlpr:
So, I tested all the upcoming weeks currently selectable. Everything is as expected.

sircharlo:
Seems to works well! One minor glitch I noticed: now, the inline videos are attempted to be fetched and fail, since there is no JWPUB for them: 

image

Perhaps internal refs that have the `PublicationCategorySymbol` of `web` (from the `RefPublication` table) should be ignored? From what I can tell, that category doesn't have JWPUB files made available anyways.

mtdvlpr:
Hmm, this also happens on the master branch, so it's unrelated to this new piece of code

mtdvlpr:
So the Russian workbook refers to the videos used in the ministry using a non existing publication. I though it was just a mistake, but they're using it consistently so it must have a reason. I think we can indeed exclude web extracts to avoid these calls. I'll include it in this branch

mtdvlpr:
If that's all, we can merge this now :)

mtdvlpr:
Since this issue is also present in the upcoming midweek meeting (the same Ahab chart is referenced), we might consider releasing v22.12.0 early.

sircharlo:
> Since this issue is also present in the upcoming midweek meeting (the same Ahab chart is referenced), we might consider releasing v22.12.0 early.

Definitely! Good detective work 🙂

sircharlo:
Do you think 22.12.0 is ready for primetime?

mtdvlpr:
> Do you think 22.12.0 is ready for primetime?

Yes! You can complete your translations and then we're good to go, I'd say :)

sircharlo:
Done 😃

mtdvlpr:
Then the honor is all yours ;)

PR #795, opened 2022-12-01
chore: log error for failed api call

mtdvlpr:
@sircharlo, is this what you intended with your Sentry comment?

### DIFF
--- a/src/renderer/plugins/media.ts
+++ b/src/renderer/plugins/media.ts
@@ -519,6 +519,7 @@ const plugin: Plugin = (
         })
       } catch (e: unknown) {
         $log.debug(params, mediaItem)
+        $log.error(e)
         try {
           result = await $pubMedia.get('', {
             params: {


sircharlo (approved):
Looks good! Maybe finally the reason for those extra calls will become clear..!

mtdvlpr:
I've also been thinking. Those extra fallback calls. What's the use case for them? I've never seen a scenario where the initial call fails and the fallback succeeds (except the sjj songbook, but that's now correctly done in the initial call)

sircharlo:
I believe it was for sign languages. 

Compare [this call](https://b.jw-cdn.org/apis/pub-media/GETPUBMEDIALINKS?output=json&pub=sjjm&track=81&langwritten=ASL) with [this one](https://b.jw-cdn.org/apis/pub-media/GETPUBMEDIALINKS?output=json&pub=sjj&track=81&langwritten=ASL).

In the JWPUB database for `ASL` for example, the `KeySymbol` is still `sjjm`: 
image

However, fetching that publication fails, since for some reason the "m" is dropped for sign language media in some cases. This is why [this](https://github.com/sircharlo/meeting-media-manager/blob/e4576ebe33d0a6a71dbc21e0bd214af4447b891c/src/renderer/plugins/media.ts#L546) exists.

As for this, this is indeed for old songbooks, like for the Armenian West. In the JWPUB database for `R`, the `KeySymbol` is `iasn`: 
image

However, compare [this call](https://b.jw-cdn.org/apis/pub-media/GETPUBMEDIALINKS?output=json&pub=iasn&track=104&langwritten=R) with [this one](https://b.jw-cdn.org/apis/pub-media/GETPUBMEDIALINKS?output=json&pub=iasnm&track=104&langwritten=R). The extra letter needs to be added.

Interestingly, WOL seems to do a similar "fetch all" in most cases:
image
image

mtdvlpr:
Alright, then I think we can remove the last one, since the sign language is handled here: 

https://github.com/sircharlo/meeting-media-manager/blob/e4576ebe33d0a6a71dbc21e0bd214af4447b891c/src/renderer/plugins/media.ts#L492-L498

Issue #816, opened 2022-12-04
[Feature Request] Secondary media language as fallback


Stephensce:
### Full error log
```
{
  "1670174954341": [
    "No WE meeting data found for 2022-12-04!"
  ]
}
```
### Error details
```
{
  "stack": "Error: No WE meeting data found for 2022-12-04!\n    at f.<anonymous> (app://./_nuxt/b10be0e.js:1:267052)\n    at async f.getWeMedia (app://./_nuxt/eaf35a3.js:1:7273)\n    at async Promise.allSettled (index 1)\n    at async f.startMediaSync (app://./_nuxt/eaf35a3.js:1:9171)\n    at async f.execute (app://./_nuxt/eaf35a3.js:1:4342)",
  "message": "No WE meeting data found for 2022-12-04!",
  "__sentry_captured__": true
}
```

mtdvlpr:
It seems the Study Edition of The Watchtower is not available in Irish. Or am I mistaken?

mtdvlpr:
There are only questions if I look here:
https://www.jw.org/en/library/magazines/?contentLanguageFilter=ga&pubFilter=w&yearFilter=

Stephensce:
Hi
Thank you for your reply. So the question I was going to ask is, about an
unusual situation as we do a hybrid meeting in both English and Irish. The
items the publications and videos are getting more available in Irish but
until then we have to use both languages to fill the gaps if you know what
I mean, so we were wondering is it possible to pull down the media for each
meeting in both English & Irish? I noticed thet M3 will display the year
text in Irish on the 2nd screen which is fantastic, if we could pull
media/Videos and songs for each meeting in both languages while using the
Irish profile & years text that would be amazing.
Thanks so much for your work so far on the app.
It is a very useful tool.

Kind Regards
Stephen.

mtdvlpr:
That is definitely an interesting request! I believe there is no media (Watchtower  meeting workbook or kingdom song videos) yet in Irish, so implementing this would currently only allow you to have English media with an Irish yeartext. Am I correct?

Stephensce:
Yes it's unusual but the brothers are working hard on getting more
translated, For the Wt we read the paragraphs in English, ask the
questions in Irish & translate our answesr which we give as much as
possible in Irish.The full watchtower study is not available in Irish, a
good number of the videos on the website are now available in Irish, the
first call and return visits videos are available. There are now
around 55 irish songs translated with the lyrics available,the Congregation
[...]

PR #817, opened 2022-12-04
feat: add option for fallback language

mtdvlpr:
Related to #816

@sircharlo, since this is quite a big modification to a big part of media fetching logic, I want two sets of eyes looking over it👀

TODO: 

- [x] use media fallback language as yeartext fallback
- [x] use availability check to prevent failing api calls for mwb and w pubs


sircharlo:
For sure; I like the idea of this feature. I guess the way such a feature should work is:

    Fetch mwb in primary language
    IF SUCCESS: 
        Fetch media as per usual (using language as defined in mwb)
    IF FAIL:
        Fetch mwb in fallback language
        Fetch media in **primary** language 
            IF FAIL: Fetch media as per usual (using language as defined in mwb)

And the same for the w...

mtdvlpr:
That is almost how I've implemented it! I don't know if the mwb is of the primary language, so I always try to fetch media using the primary language and if that fails the specified language in mwb or if that is the same as the primary language, the fallback language.

sircharlo:
> That is almost how I've implemented it! I don't know if the publication is of the primary language, so I always try to fetch media using the primary language (when fallback lang is specified) and if that fails the specified language in the publication or if that is the same as the primary language, the fallback language.

Excellent, that seems like airtight logic. Will need to be tested with a few scenarios, of course, but the logic is good.

mtdvlpr:
> Excellent, that seems like airtight logic. Will need to be tested with a few scenarios, of course, but the logic is good.

Yeah, the different sets of scenarios (I think) are:
- [ ] Available primary language without fallback (e.g. English): should function the same
- [ ] Available primary language with fallback (e.g. English with Dutch): should **NOT** fetch anything for the fallback language
- [ ] Unavailable primary language without fallback (e.g. Irish): should function the same
- [ ] Unavailable primary language with fallback (e.g. Irish with English): Should use primary language whenever possible
- [ ] Semi-available primary language without fallback (e.g. Abui): should function the same (use fallback specified in mwb)
- [ ] Semi-available primary language with fallback (e.g. Abui with English): use fallback specified in mwb over custom fallback

sircharlo:
I came across an error with a clean media cache, ABU as main and E as fallback: 

image

mtdvlpr:
Ah, I missed it because I still had English publication cache. The fallback language cache should also be removed when clearing cache

sircharlo:
Also, ABU as main with no fallback gives errors fetching songs, even though the songs exist online:

image

sircharlo:
> I noticed a mistake when trying to fetch the Abui media without a fallback. The video 'Jehovah Remembers Our Hard Work' was fetched for ABU, which failed. JW.org fetches the video in the fallback language IN. However, I could not find a way to extract the language of this video from the jwpub db, since it is not an Extract which has a Link attribute.

As for this, it might work on the website, but look at what happens in JW Library itself:

image

If JW Library itself doesn't do it, I don't see how we can. No doubt the website is using some other internal database, but JW Library uses JWPUB files, like we do. If they add the feature in JWPUB files to specify a language for the `Multimedia` table, then we can talk 😉

sircharlo:
On a side note: look at this [article](https://wol.jw.org/abz/wol/d/r278/lp-abu/202022413). They forgot to make links into links? 😆

mtdvlpr:
> > I noticed a mistake when trying to fetch the Abui media without a fallback. The video 'Jehovah Remembers Our Hard Work' was fetched for ABU, which failed. JW.org fetches the video in the fallback language IN. However, I could not find a way to extract the language of this video from the jwpub db, since it is not an Extract which has a Link attribute.
> 
> As for this, it might work on the website, but look at what happens in JW Library itself:
> 
> image
> 
> If JW Library itself doesn't do it, I don't see how we can. No doubt the website is using some other internal database, but JW Library uses JWPUB files, like we do. If they add the feature in JWPUB files to specify a language for the `Multimedia` table, then we can talk wink
> 
> No doubt an ABU congregation could simply use ABU as main language, and IN as fallback in the settings, to work around this current limitation of JWPUB files.

That's good to know :)

mtdvlpr:
@sircharlo, the bugs you found should be resolved now. I made a mistake in the new language detection when using $pubPath(). It now tries mediaLang first, if that doesn't exist, lang specified in mwb and then fallback lang. It skipped straight to fallbackLang even if there is none previously

sircharlo:
> @sircharlo, the bugs you found should be resolved now. I made a mistake in the new language detection when using $pubPath(). It now tries mediaLang first, if that doesn't exist, lang specified in mwb and then fallback lang. It skipped straight to fallbackLang even if there is none previously

They are, I was just testing it out. Works great for ABU (minus that one video), ABU/IN, and GC/E.

2 ideas: 

1. When the `mwb` and/or `w` is not available in a language, perhaps the notification could be modified to also say something about "perhaps consider setting a fallback language"?
2. Perhaps the JW tile could indicate the fallback language alongside the main language, when present? For example, right now it isn't clear that GC is configured along with W as a fallback:

image

mtdvlpr:
Done and done :)

sircharlo (approved):
🎉 🎉 🎉

mtdvlpr:
> No time like the present smile

Alight! Maybe give the translators a few days to update their translations for the fallback strings and then release a 12.1?

sircharlo:
> > No time like the present smile
> 
> Alright! Maybe give the translators a few days to update their translations for the fallback strings and then release a 12.1? Some of them are very fast when it comes to new translation strings. Btw, will they receive a notification for the updated unavailable strings? Those were already translated by them, but should be changed now.

Good idea, and I don't think so. I'll check when this merges, but I don't recall ever being notified of new translations pending. I'd love it if GitLocalize had email notifications..!

sircharlo:
Hey! It works, but no notification. So translators will have to check back manually.

image

Issue #818, opened 2022-12-05
[ENHANCEMENT] Button labels can be too long


sircharlo:
**Describe the bug**
Depending on the language, some button labels can be too long.

**To Reproduce**

1. Set app language to West Pidgin
2. Observe the "clear logs" button; the "home" button gets cut off

**Expected behavior**
Perhaps a max-width should be set for buttons with dynamically loaded text, with auto-clipping and in-element scrolling of the text? Or simpler, perhaps buttons should be all be represented by icons with hover text (example: the "sort media" button).

**Screenshots**
![image](https://user-images.githubusercontent.com/1275666/205670247-6d6dbc06-66e9-45dc-a157-e08d0df612d4.png)

mtdvlpr:
I would go for icon buttons with translated hover text

sircharlo:
Fixed in #819

PR #819, opened 2022-12-05
fix: make report and clear cache icon buttons

mtdvlpr:
Relates to #818

@sircharlo, what do you think of this?


sircharlo:
Awesome! The only other button that I think could benefit from this is the "download all songs" button.

sircharlo (approved):
Awesome! Much cleaner now.

Issue #839, opened 2022-12-11
[Enhancement] choose which jwpub images to add


mtdvlpr:
When using a jwpub file to add images to a certain date, whether it be from a public talk or a lesson in the Enjoy life forever book, once you select the public talk or lesson, you are forced to add all images/videos that are in that public talk or lesson.

It would be nice if you could remove some of those images/videos before pressing save. This saves time  because those files won't have to be uploaded to the congregation server to be removed right after again.

This could be extended to the scenario when you try to upload multiple files at once, but change your mind about one of them. Instead of having to cancel the whole process, you could just cancel that one file

mtdvlpr:
@sircharlo, thoughts?

sircharlo:
> @sircharlo, thoughts?

Love love love it

PR #840, opened 2022-12-11
feat: ignore single file in media list

mtdvlpr:
When adding local files or files extracted from local jwpub files, you can prevent a single file from being added by clicking it. It will get a strike-through, similar to the hide feature. Clicking it again will remove the strike-through. When you press save, only the files without a strike-through will actually be added.

Related to #839


sircharlo (approved):
Awesome idea! To prevent time wasting; that’s always a good thing 😄

PR #857, opened 2022-12-15
feat: song picker in presentation mode

mtdvlpr:
It works as follows:
- Click music icon in media controls in presentation mode to toggle song picker
- Select a song and it gets added to the top of the media list
- If the song already exists in the cache, it will use that local file, otherwise it will stream the song and have a download option
- After downloading the song, the local cache file will be used instead of streaming it
- In both scenarios, the song can be paused and scrubbed and custom start and stop times can be set
- The song can **NOT** be moved by enabling the sort feature and **cannot** be played/stopped using the keyboard shortcuts

Related to #845


sircharlo (commented):
I like it very much! 
Preliminary feedback or questions: 
- would the user benefit from seeing a sort of delineation between the added song and the rest of the media? (Different background color, a border, or something along those lines)
- would there be a benefit to starting the download automatically and hiding the download button? (If user click stream before download is finished, play from remote URL nonetheless; otherwise of course play from cache)
- I would add some mouseover text to the new button, something like "Quickly add a song to the media list"
- I would add the "fa-plus" icon to the new button, to stay consistent with the other icons. ("song + shuffle", "song + add")

mtdvlpr:
Great points! It is already different, because it's in a seperate list, so it's in its own box. We could make it clearer with a horizontal line perhaps. Auto download might be useful, I'll have to test what happens if a download is completed while streaming the song.

sircharlo (approved):
Man. This app is starting to be really something, isn't it. You're doing an awesome job. Thanks so much for your time and work on this!

Issue #871, opened 2022-12-18
[BUG] Media added from JWPUB import is in wrong language


sircharlo:
**Describe the bug**
Media added from JWPUB import is in wrong language. The language is fetched in the language of the congregation instead of in the language defined in the JWPUB file.

**To Reproduce**

1. Set congregation language to English
2. Add JWPUB file of other language that has some videos
3. Videos are downloaded in English rather than in the language of the JWPUB file

**Expected behavior**
Media added from JWPUB import should be in the language of the JWPUB file.

PR #872, opened 2022-12-18
Fix/several fixes

mtdvlpr:
@sircharlo, I made it a draft, in case you wanna test it yourself :)

Turns out, for some reason there was some duplication in the code. The multimedia is fetched, which also fetches the videos. And then for each video, the video is fetched again and the second result (without lang checking) is used. I removed the redundant fetch and now also use FilePath as a way to extract the language of the media file (with a check against langs.json).

Besides that, the videos didn't make use of the title property so they had the FilePath as safeName. That's fixed and when invalid settings are fixed, the tab for those settings (application, media, meeting, cong) is automatically closed. I now check if you're on a specific settings tab before closing (otherwise you wouldn't be able to open it again)

Related to #871, #873


sircharlo:
You don't need to add URL and pull the printed page from the web!

sircharlo:
They're already available in the JWPUB file.

mtdvlpr:
> [Here](https://github.com/sircharlo/meeting-media-manager/blob/695941a470c8989a9ab7f1d7f626b2434324520b/src/renderer/plugins/media.ts#L465) is where the SVG files are specifically excluded (`Multimedia.CategoryType <> 6`). You can comment it out to see it at work immediately.

Alright, so what do you think? Include them, don't, give the option to?

sircharlo:
> Alright, so what do you think? Include them, don't, give the option to?

Maybe a god solution would be to add an option, disabled by default, in settings? Something like "Include media for both the printed and digital editions for publications that support this"?

mtdvlpr:
@sircharlo, let me know when you've tested the JWPUB importing another language and are happy with this PR :)

sircharlo:
Very nice! One small bug i noticed:

1. Clear cache
2. Click on non-meeting day
3. Add media from document with pictures and videos from JWPUB file of language other than congregation
4. (Successful so far)
5. Click media sync button
6. Errors about extracting and processing the publication that was manually imported!
7. Restart app
8. Click media sync button
9. Errors are gone

Seems to be an issue var variable reuse/reinitializtion, as if the JWPUB file gets added to the media to be processed during a normal MW/WE/cong. sync.

mtdvlpr:
Hmm, in this situation I really miss my Vue Devtools (it stopped working for some reason)

mtdvlpr:
Oh, wait, I think I found it! We're saving the databases we extract from jwpub files in the store under a pub and issue key. So if the same pub and issue want their database, we give them the database previously extracted (not accounting for a possible different language)

mtdvlpr:
@sircharlo, it should work correctly now :)

sircharlo:
> @sircharlo, it should work correctly now :)

Works great! I hate that we can't "logic" away those crown icons, but they're indistinguishable for other pictures in the db. 🤷 
I like the settings toggle, super polished.

sircharlo (approved):

Issue #874, opened 2022-12-19
[BUG] lff media is included for 5-min bible studies


mtdvlpr:
**Describe the bug**
When a bible study for the apply yourself portion of the midweek meeting includes images/videos in the referenced lff lesson, they are included in the media list. Since the instructions specify that such videos are not to be played during the assignment but merely referenced, they should preferably not be included in the media list.

**To Reproduce**

1. Go to 'December 26 - January 1'
2. Click on 'Fetch media'
3. Do 'Open presentation mode'
4. See 'Jehovah has done so much for me' video

**Expected behavior**
The media referenced for a 5-min bible study should be ignored

**System specs (please complete the following information):**

- OS: All
- App Version: 22.12.1

**Additional context**
@sircharlo, since you have more knowledge about the fetching logic, maybe you know an easy fix for this. The only solution I can think of is comparing the paragraph numbers of all referenced lff media and only using the media of the latest paragraph (since that's the congregation bible study).

sircharlo:
There it is! That's why I had disabled it in the previous version, now I remember. 
Related to #744 and #745. 
There's no easy fix here from a "logic" point of view here from what I can see. Media from refs can either be processed (like what was needed in #744) or not (like for the 5 min. talk during the week of Dec. 26-Jan 1st).
The only thing I can think of would be to ignore _videos_ for ref media (as videos to be shown during a meeting will always be linked to "directly", not from refs), while processing _images_ normally (sometimes, images from refs are used by speakers/students and can contain useful information).

mtdvlpr:
I'm confused. Isn't it only supposed to do that for internal refs of internal refs. The lff book isn't an internal ref, right?

sircharlo:
No, **I'm** confused. 😕 
#744 was _[checks]_ media from a `Document` referred to from an `InternalLink`.
This current issue is _[adjusts glasses]_ a video being fetched from a `Document` referred to from the `Extract` table. The problem is that this is exactly the same was that the CBS media gets fetched.

sircharlo:
Programmatically, there's no way to distinguish from `Extract` numbers 76 and 79...

![image](https://user-images.githubusercontent.com/1275666/208487608-4e940648-d215-41e0-8a87-b6c58f5b2cea.png)

mtdvlpr:
[...]

PR #879, opened 2022-12-19
fix: ignore lff videos other then cbs

mtdvlpr:
@sircharlo, I suppose this does the trick. Ignore videos from Extract when another lff extract is present with a greater BeginParagraphOrdinal.

Related to #874


sircharlo:
Would this cause issues for languages that don't study the `lff` book? Or for example in the future when the publication changes. Just thinking out loud. I tend to dislike any publication hardcoding whenever possible.
(I wish there was an clean way/API to fetch the appropriate songbook for each language lol)

mtdvlpr:
> Would this cause issues for languages that don't study the `lff` book? Or for example in the future when the publication changes. Just thinking out loud. I tend to dislike any publication hardcoding whenever possible. (I wish there was an clean way/API to fetch the appropriate songbook for each language lol)

It only changes the logic if the extract is of pub lff and there's another extract of lff. Otherwise the normal logic for fetching is used.

mtdvlpr:
I am wondering though. Weren't the exclude lffi images and exclude lffi videos toggles originally meant for this use case?

sircharlo:
> @sircharlo, I suppose this does the trick. Ignore videos from Extract when another lff extract is present with a greater BeginParagraphOrdinal.

Would this logic be problematic in the (unlikely) case that there would be multiple extracts with the same BeginParagraphOrdinal? (Example, the CBS is based on paragraphs, and a chart found later in the same document. This is represented as two seperate lines in the Extract table.)

sircharlo:
> I am wondering though. Weren't the exclude lffi images and exclude lffi videos toggles originally meant for this use case? Should we repurpose them to exclude the images / videos of the lff pub when used outside of cbs?

But to answer your question, I suppose that those toggles could indeed be repurposed. I believe that "exclude images from lffi" is no longer needed imo, and "exclude audio/video from lffi" could become something like "Include audio/video files from the Live Life Forever publications for the CBS only"

mtdvlpr:
> Would this logic be problematic in the (unlikely) case that there would be multiple extracts with the same BeginParagraphOrdinal? (Example, the CBS is based on paragraphs, and a chart found later in the same document. This is represented as two seperate lines in the Extract table.)

No, because then there would not be a match with a greater BeginParagraphOrdinal, so the normal fetch logic would be used and everything would therefore be included

mtdvlpr:
> But to answer your question, I suppose that those toggles could indeed be repurposed. I believe that "exclude images from lffi" is no longer needed imo, and "exclude audio/video from lffi" could become something like "Include audio/video files from the Live Life Forever publications for the CBS only"

Based on your other comment, I would say that the exclude images can be repurposed to exclude images from lff outside of cbs (same use case as before, just updated to lff because we are past the first 3 lessons) and the exclude videos can be removed (seeing as the videos are never to be played). Am I wrong?

sircharlo:
I just haven't heard from anyone actually using the toggle to disable _images_ from the lffi brochure, seing as the images are relevant and often used by the students. I wonder if this toggle is necessary at all.
And since the videos won't be downloaded at all with this PR, then for sure, the "exclude audio/video files" toggle won't be necessary anymore, unless at some point the lessons from the brochure are used for the initial and second conversation demonstrations.

mtdvlpr:
Our congregation never uses those images, so I would like the toggle for myself then :)

mtdvlpr:
As for the videos, even for initial and return visits, the video always has the description: reference the video, but don't play it.

mtdvlpr:
@sircharlo, that should do the trick :)

P.S. To make it work locally, you need to set the migration v23.1.0 to v22.0.0 and in your prefs.json you need to set the migrated version to before v21.0.0 because locally the Electron version is used (22.0.0) instead of the package.json version.

mtdvlpr:
> unless at some point the lessons from the brochure are used for the initial and second conversation demonstrations.

I just read it correctly. For that use case, we could leave the lffi logic in, force the video filtering and use the image toggle for the brochure as well.

sircharlo:
> @sircharlo, that should do the trick :)
> 
> P.S. To make the migration work locally, you need to set the migration v23.1.0 to v22.0.0 and in your prefs.json you need to set the migrated version to v21.0.0 because locally the Electron version is used (22.0.0) instead of the package.json version.

Looks awesome. Picture from student assignment shows when toggle is disabled, and vice versa.

sircharlo (approved):

PR #888, opened 2022-12-20
ci(Mergify): configuration update

sircharlo:
This change has been made by @sircharlo from Mergify config editor.


sircharlo:
@mtdvlpr Laziness wins again.

Issue #892, opened 2022-12-20
[Enhancement] Make the tooltip persistent for "click again to confirm" tooltips


sircharlo:
**Describe the feature**
When a user clicks on something that requires a second click (like stopping a video, clearing the cache, or starting the background music), the tooltip shows up to indicate to click again. However, if the user's mouse moves away from the button, the tooltip disappears, making it hard for the user to realize what is happening, or what they need to do.

**How would this feature work?**
It would be nice if the tooltip could persist until the button goes back to its 'unclicked' state.

**Who would benefit from it?**
Everyone who clicks on things and moves the mouse fast without reading, like me.

**Additional context**
I tried to play around with the shuffle button to make the `v-tooltip` persistent using `v-model` as per the [docs](https://vuetifyjs.com/en/components/tooltips/#visibility), but I couldn't get `v-model` to play along with the `v-if`. I must be missing something. I [got around it](https://github.com/sircharlo/meeting-media-manager/commit/168d9b45e687222d7d67a2f2f8e60f2c77c772c7) using [`close-delay`](https://vuetifyjs.com/en/api/v-tooltip/#props-close-delay), but maybe that's not the best way...

mtdvlpr:
@sircharlo, see #893 :)

sircharlo:
> @sircharlo, if you look at mine (#879), I'll look at yours :)

Sorry, I had to leave quickly! 😆 I'll take a look for sure 😉

PR #893, opened 2022-12-20
feat: make click again tooltips persistent

mtdvlpr:
@sircharlo, you also needed the disabled prop to prevent it from changing its state internally

Related to #892


sircharlo:
> @sircharlo, you also needed the disabled prop to prevent it from changing its state internally
> 
> Related to #892

Wow, so simple. Is it just me or is that not in the [docs](https://vuetifyjs.com/en/components/tooltips/#visibility), or even in the example on the docs page?

sircharlo (changes requested):
Looking good, except for the "clear cache" button tooltip

mtdvlpr:
@sircharlo, I haven't found a way to fix this. Might I suggest combining the labels into one? `clear text (click again to confirm)`. The tooltip would still not show after clicking, but at least the click again text can be read beforehand.

sircharlo:
> @sircharlo, I haven't found a way to fix this. Might I suggest combining the labels into one? `clear text (click again to confirm)`. The tooltip would still not show after clicking, but at least the click again text can be read beforehand.

And why is the same solution you used previously not working now (disabled + value)? I tried to do something along the lines of `:disabled="color !== warning" :value="color !== warning"`, but that didn't work naturally. I don't get why it works in other places but not here. Is it because the text of the tooltip is being replaced dynamically, therefore resetting some property somewhere?

mtdvlpr:
It's because of the tooltip that's already there. I don't understand it fully, but the first tooltip is preventing the second tooltip from becoming active right after the click.

mtdvlpr:
@sircharlo, I managed to improve the clear cache tooltip! It will now immediately show the click again tooltip when clicking on the button instead of having to move away from and back to the button. Haven't gotten the persistent working (yet), but after digging through the API docs I found some potential fixes.

mtdvlpr:
@sircharlo, got it! I removed the activator on and attrs, which make the tooltip responsive to hover, focus and click. Doing that will also make the disabled prop unncessary, so I simplified all other persistent tooltips as well.

sircharlo:
> @sircharlo, got it! I removed the activator on and attrs, which make the tooltip responsive to hover, focus and click. Doing that will also make the disabled prop unnecessary, so I simplified all other persistent tooltips as well.

Works great for me, at least visually! However the console throws this error now for the shuffle button, can that be ignored safely?

image

mtdvlpr:
@sircharlo, fixed :)

sircharlo (approved):
Awesome sauce

PR #915, opened 2022-12-23
Chore/several fixes

mtdvlpr:
Firstly, I removed the `needs: test` from the ci, because now the build jobs are waiting for the test jobs to finish, which makes the entire workflow take twice as long. The test job runs the tests on a single build and the build job tries to build all supported builds (win32 and win64, appImage and Debian). These can run simultaneously. If you wanted to prevent a release with failing tests, that's basically already done by running the tests against every pull request.

Secondly, I'd like your opinion on the issue template forms I've created. Template forms allow more customization and makes the issue template nicer to work with. We could keep the regular templates and have the two versions live side-by-side for the user to choose what they prefer or replace the current templates with the forms altogether.


sircharlo:
> Firstly, I removed the `needs: test` from the ci, because now the build jobs are waiting for the test jobs to finish, which makes the entire workflow take twice as long. The test job runs the tests on a single build and the build job tries to build all supported builds (win32 and win64, appImage and Debian). These can run simultaneously. If you wanted to prevent a release with failing tests, that's basically already done by running the tests against every pull request.

Isn't it just less wasteful in terms of computing resources to wait for the test to succeed before doing the build step? We know the build is useless if the test didn't succeed, so why build at all if the test failed?

sircharlo:
> Secondly, I'd like your opinion on the issue template forms I've created. Template forms allow more customization and makes the issue template nicer to work with. We could keep the regular templates and have the two versions live side-by-side for the user to choose what they prefer or replace the current templates with the forms altogether.

Oh yeah I love this! So much cleaner!! 💯

mtdvlpr:
> Isn't it just less wasteful in terms of computing resources to wait for the test to succeed before doing the build step? We know the build is useless if the test didn't succeed, so why build at all if the test failed?

True, but you also don't want to wait 20 minutes for the workflow to succeed, that's wasting time. Ideally the whole workflow would fail as soon as either a test job or a build job fails. I haven't been able to find such a setting though.

mtdvlpr:
> Oh yeah I love this! So much cleaner!! 100

Based on you answer, I guess you want to replace the old ones?

mtdvlpr:
> > Isn't it just less wasteful in terms of computing resources to wait for the test to succeed before doing the build step? We know the build is useless if the test didn't succeed, so why build at all if the test failed?
> 
> True, but you also don't want to wait 20 minutes for the workflow to succeed, that's wasting time. Ideally the whole workflow would fail and stop as soon as either a test job or a build job fails. I haven't been able to find such a setting though.

@sircharlo, this might be what we're looking for:
https://github.com/vishnudxb/cancel-workflow

sircharlo:
> > > Isn't it just less wasteful in terms of computing resources to wait for the test to succeed before doing the build step? We know the build is useless if the test didn't succeed, so why build at all if the test failed?
> > 
> > 
> > True, but you also don't want to wait 20 minutes for the workflow to succeed, that's wasting time. Ideally the whole workflow would fail and stop as soon as either a test job or a build job fails. I haven't been able to find such a setting though.
> 
> @sircharlo, this might be what we're looking for: https://github.com/vishnudxb/cancel-workflow

We're both searching at the same time 😆 Check out my latest commit for a potential workaround, based on [this](https://stackoverflow.com/a/73225704).

mtdvlpr:
It worked!

sircharlo:
> It worked!

image

Indeed!

sircharlo:
> > Oh yeah I love this! So much cleaner!! 100
> 
> Based on your answer, I guess you want to replace the old ones?

Why not. Any scenario where the old ones would be better? I think not.

mtdvlpr:
I'm keeping the app_bug_report (for now) until we have tested if you can open an issue from a url using an issue form

Issue #682, opened 2022-11-14
[BUG] Renaming a file sometimes returns a 404 from the cong server


mtdvlpr:
**Describe the bug**
When renaming a file in the manage media screen, a 404 error is often thrown.

**To Reproduce**

1. Go to 'a specific day'
2. Click on 'rename file'
3. Do 'Rename a file'

**Expected behavior**

**Logs**
See Sentry for logs

**System specs (please complete the following information):**

- OS: [e.g. Windows 10]
- App Version [e.g. 22.11.0]

**Additional context**
See Sentry for more details

sircharlo:
Hmm. I can't reproduce this, try as I might. Odd.

mtdvlpr:
Yeah, I have the same issue. That's what makes this so difficult to tackle. There is something wrong though, since multiple users are experiencing this at times.

sircharlo:
Hmm. Check my comment in Sentry.

mtdvlpr:
@sircharlo, I'm closing this issue until it appears again in the latest release...

mtdvlpr:
Unfortunately, the issues remains...

mtdvlpr:
I think we need an actual user to experience this and report their findings before being able to solve it...

sircharlo:
> I think we need an actual user to experience this and report their findings before being able to solve it...

You're probably right; it's definitely an intermittent issue at worst.

mtdvlpr:
Definitely, otherwise it would've been reported already

PR #927, opened 2022-12-23
Fix/several fixes

mtdvlpr:
@sircharlo, I think we finally got him! I was just shown a video of a user trying to rename a file. He misclicked and clicked on the filename and then clicked on the pencil icon to rename. But that second click activated the delete function. Now the delete and hide actions are only activated when actually clicking on the icon and not on the filename.

Related to #682


sircharlo:
Oh wow! Awesome catch!!

Issue #993, opened 2023-01-01
[Bug]: Unable to "auto-stop" pre-meeting music if meeting start is in more than 1 hour


sircharlo:
### What happened?

If the meeting start is in more than 1 hour (for example, 61 minutes), the music stops prematurely.

### To Reproduce

1. Set meeting music auto stop to "15 seconds before start of meeting"
2. Set meeting start time to in 61 minutes
3. Start meeting music
4. Observe that the music countdown timer does not display something like "60:14", but rather "00:14"
5. Music stops when timer reaches zero
6. Music can be restarted at this point, and the timer is correct ("59:53" etc.)

### What did you expect to happen?

I expect to be able to start the music many hours in advance if needed. (Edge case, but it's happened in the past.) 

### Version

v23.1.0

### What platform are you seeing the problem on?

Windows

PR #996, opened 2023-01-02
fix: auto stop shuffle if started more than 1 hour in advance

mtdvlpr:
## Proposed Changes

If shuffle music starts more than 1 hour before the meeting, the minutes displayed will be the correct number of minutes left (60+).

Related to #993


sircharlo (approved):
Works great!

Issue #990, opened 2022-12-31
[Feature]: Shared Media Cache


danielengels:
### Is your feature request related to a problem?

We share our hall with another assembly. Everyone has their own Windows account because we do different things with it. Unfortunately our internet connection is very bad. So downloading the media is quite time-consuming. So far, we have often brought the media with us from home.
Both assemblies now want to use the Meeting Media Manager. Is it somehow possible to share the media cache. Then one meeting would benefit from the media cache of the other meeting. 

### Describe the solution you'd like

It would be good to select the media cache folder to have it in one place where both windows users can use it? Maybe like the media folder for presenting the media

### Describe alternatives you've considered

Maybe we could use one windows user again - but there a some things we do different and no one, want to change settings from the other.

mtdvlpr:
Hi @danielengels, I have not tried this myself yet, but have you tried installing M³ system-wide instead of only for the current user? I would imagine that the application (along with the cache) would be installed in a shared location.

danielengels:
Hi @mtdvlpr,
I play a little arround - this option only change, where the software itself is installed. On Windows this changes from Program Files (system-wide) to %appdata%/local/programs (user installation). And you don't need admin rights, when you don't install system-wide but this depends on the installation path.

mtdvlpr:
@danielengels, alright, thanks for trying. In that case we can definitely add an option to overwrite the default location for publication cache! Will keep you posted here.

mtdvlpr:
Hi @danielengels, I have a version for you to test out. You can find it [here](https://drive.google.com/drive/folders/1_2CzKieYGaK3LcYrEwzNQgfTpNReOSx9?usp=sharing). You can set a custom cache location in the settings which will migrate the existing cache to the new location.

danielengels:
Hi Manoah,
seems to work quiet well. But there are two things I don't know what happens.
1. The download is much faster, as they uses the same cache. But at the end of the new download each congregation have to download something. But it is still much much less then before. To reproduce: create two congregation (i do in the same user profile. Start downloading for the first congreagation. This takes a while. After closing and opening MMM with the other congregation u see that it don't need so long. But at "the end" (maybe it is not) it seems like there is something downloaded. After closeing and opening again and refreshing the media this download don't happens again.
[...]

PR #997, opened 2023-01-02
feat: allow custom cache location

mtdvlpr:
## Proposed Changes

A new setting `custom cache location` which would become the location where the Publications and Fonts folders are stored.

Related to #990


sircharlo (approved):
Nice feature! Works very well and will be super useful.

sircharlo:
@mtdvlpr Think I found a bug with this. If I set a custom cache location, relaunch the app, and go to Settings, every time I get two warnings:

image
image

mtdvlpr:
It' because the settings are loaded in, so the cache path goes from null to your path. Then it tries to move your cache from the default path to your "new" custom path. The problem is it can't be differentiated from a manual change. The error in the console is another matter. The identifier should be included in the message id.

mtdvlpr:
@sircharlo, fixed in #1021

Issue #382, opened 2022-05-31
Unable to select media presentation window in Zoom on macOS


rickinhissmart:
<img width="1863" alt="Screenshot  " src="https://user-images.githubusercontent.com/14828029/171295101-5871e8d3-844b-430d-b4e7-e8e0d3a1f00c.png">
hello,many thanks for hard work improving your app

zoom does not seem able to share screen of your app ie (set as in m3 second monitor or  in a separate window, only giving option for settings window to share.

spec, MacOS Monterey
MacBook Pro  2021
apple m1 pro
16 gb mem
zoom

sircharlo:
I can't reproduce this on Windows:
![image](https://user-images.githubusercontent.com/1275666/171308130-4ed56764-e5ef-486d-80d3-a851af041c09.png)

I'll try to play around a bit and see if I can figure something out.

rickinhissmart:
yes you can send test build and will try for you

rickinhissmart:
Do you have a test beta to try in mac ?

sircharlo:
Try this alpha dev package: https://we.tl/t-Y6BtQMuj3u

Note that it is not functional and only to be used for debugging.
It should open a maximum of 4 media windows. They might not all work, but that's ok. We just want to see what Zoom sees.
Please let me know the result in the Zoom sharing window. 

For example, on Windows, this is what I see: 

![image](https://user-images.githubusercontent.com/1275666/172435069-f1b0d96b-3481-4d25-8441-284b88c3a702.png)

rickinhissmart:
<img width="1983" alt="Screenshot 2022-06-07 at 17 57 43" src="https://user-images.githubusercontent.com/14828029/172483618-8cf1279d-9e4c-405e-bc6f-dc569d2615c6.png">

does this help, kind regards

sircharlo:
It actually does!

Try this?
https://we.tl/t-Qdw9WX3Zpr

rickinhissmart:
<img width="1980" alt="Screenshot 2022-06-07 at 23 15 47" src="https://user-images.githubusercontent.com/14828029/172492317-e77d6b64-a45b-4848-aa31-7d00b2975ade.png">

still no presentation window

rickinhissmart:
hey sircharlo  , I have also just tried on intel Mac running Big Sur 
with 5.10.6 (7526) zoom 
and latest version 5.10.7 (7748).zoom
it's also having the same issue.

[...]

PR #999, opened 2023-01-02
fix: disable rounded corners if not fullscreen

mtdvlpr:
## Proposed Changes

This might be a fix for the rounded corners bug. The problem is that it does not work in combination with `fullscreen: true`. So what I do here is check if the platform is darwin and the media window changes position (fullscreen to window and vice versa) and if that's the case recreate the media window with the roundedCorners set to the value of fullscreen.

Related to #425


sircharlo (commented):
Were you able to test if this has any impact on the existing screen behavior? (Automatically put media window fullscreen when new display is added; automatically move media window to screen 1 when moving the main window to screen 2, etc)

sircharlo:
Also, were you able to test on a Mac if the workaround is successful, of it if still crashes somewhere along the way?

mtdvlpr:
> Also, were you able to test on a Mac if the workaround is successful, of it if still crashes somewhere along the way?

Not myself, but @riggles did (see #425)

mtdvlpr:
> Were you able to test if this has any impact on the existing screen behavior? (Automatically put media window fullscreen when new display is added; automatically move media window to screen 1 when moving the main window to screen 2, etc)

I haven't yet. These should also be forwarded to @riggles I think.

sircharlo:
> > Also, were you able to test on a Mac if the workaround is successful, of it if still crashes somewhere along the way?
> 
> Not myself, but @riggles did (see [#425 (comment)](https://github.com/sircharlo/meeting-media-manager/discussions/425#discussioncomment-4571645))

Ah! Very nice. I hadn't seen the new comments in that thread.

sircharlo:
> > Were you able to test if this has any impact on the existing screen behavior? (Automatically put media window fullscreen when new display is added; automatically move media window to screen 1 when moving the main window to screen 2, etc)
> 
> I haven't yet. These should also be forwarded to @riggles I think.

I agree. @riggles, would you be able to test the following scenarios on your Mac setup please? The last thing we want is for there to be a breaking change during meetings 😄 

- Start M3 with external screen disconnected. Media window should be windowed.
- Connect external screen
- Does the media window now automatically move to the new screen, fullscreen?

- If 2 screens are present, and the media window is fullscreen on one of the screens (let's call it screen 2), does moving the window with the media controls to screen 2 result in the media window automatically moving to screen 1? 

Thanks for your help!

mtdvlpr:
>     * Start M3 with external screen disconnected. Media window should be windowed.
>       Connect external screen
>       Does the media window now automatically move to the new screen, fullscreen?
> 
>     * If 2 screens are present, and the media window is fullscreen on one of the screens (let's call it screen 2), does moving the window with the media controls to screen 2 result in the media window automatically moving to screen 1?

@riggles, could you also test these scenarios when presentation mode is disabled? There's the potential that the media window is suddenly opened in these scenarios when it shouldn't be.

riggles:
@sircharlo @mtdvlpr Yes, I can test these tonight and report.

riggles:
Here is the behavior I see. Note that I am still running the version from the Google Drive link @mtdvlpr updated around 1pm EDT yesterday.

> * Start M3 with external screen disconnected. Media window should be windowed.
>   Connect external screen
>   Does the media window now automatically move to the new screen, fullscreen?

It depends on the state of the setting when the app was quit. If M3 was quit with the 2nd display connected and media set to windowed mode, restarting the app with the 2nd display disconnected and then reconnecting will not trigger a move to fullscreen on the 2nd display. The media will remain in a window on the main display. If the app was quit with 2nd display connected and in fullscreen mode, restarting the app (media will begin in a window by default) and then reconnecting the 2nd display will trigger a move to fullscreen on the 2nd display. Hope that makes sense.

> * If 2 screens are present, and the media window is fullscreen on one of the screens (let's call it screen 2), does moving the window with the media controls to screen 2 result in the media window automatically moving to screen 1?

No. The control screen and fullscreen media both stay on the 2nd display, with the control window being on top. The options in the media presentation setting change to:

- Moveable window
- Screen 1 (resolution)

Neither value is selected by default. 

> @riggles, could you also test these scenarios when presentation mode is disabled? There's the potential that the media window is suddenly opened in these scenarios when it shouldn't be.

When the media presentation mode is disabled, no media is presented on screen at all. Not when a 2nd display is connected, and not when move the control window between displays.

mtdvlpr:
@riggles, awesome work! We'll look into the scenario of moving the main window to the same screen as the media window.

mtdvlpr:
@sircharlo, we set the media window to always be on top for Windows and Linux, but not for MacOS. Do you remember the reason for that?

mtdvlpr:
> @sircharlo, we set the media window to always be on top for Windows and Linux, but not for MacOS. Do you remember the reason for that?

It seems to be changed at https://github.com/sircharlo/meeting-media-manager/commit/53fe63c7cb8823c3fdfd65e6eada743c66aff8e8

mtdvlpr:
@riggles, I have tested it on my own device (Ubuntu) and there it works as expected. As soon as the main window crosses the boundary to screen 2, the fullscreen media window is moved to screen 1. So what I've done now is add a bunch of debug statements to the existing solution. I've uploaded the version to the drive folder. If you could please go through the scenario of moving the main window to screen 2 again, use <kbd>⌘ Command</kbd> <kbd>⌥ Option</kbd> <kbd>I</kbd> to open the console and copy paste everything in there here, that would be very helpful!

sircharlo:
> @sircharlo, we set the media window to always be on top for Windows and Linux, but not for MacOS. Do you remember the reason for that? It was changed at https://github.com/sircharlo/meeting-media-manager/commit/53fe63c7cb8823c3fdfd65e6eada743c66aff8e8

In macOS, the media window could not at the time be selected for sharing if it was fullscreen. It simply didn't show up in the window list. This seemed to be a limitation of the Zoom client on macOS, more than an Electron bug. See https://github.com/sircharlo/meeting-media-manager/issues/382

riggles:
> @riggles, I have tested it on my own devices (Ubuntu and Windows 10) and there it works as expected. As soon as the main window crosses the boundary to screen 2, the fullscreen media window is moved to screen 1. So what I've done now is add a bunch of debug statements to the existing solution. I've uploaded the new version to [the drive folder](https://drive.google.com/drive/folders/1_2CzKieYGaK3LcYrEwzNQgfTpNReOSx9?usp=sharing). If you could please go through the scenario of moving the main window to screen 2 again, use ⌘ Command ⌥ Option I to open the console and copy paste everything in there here, that would be very helpful!

@mtdvlpr Here you go:

```
instrument.js:108 Object
instrument.js:108 output 2
instrument.js:108 pref screen Object
instrument.js:108 main Object
instrument.js:108 main Object
instrument.js:108 main Object
instrument.js:108 on moveMediaWindowToOtherScreen
instrument.js:108 Object
instrument.js:108 output 2
instrument.js:108 pref screen undefined
instrument.js:108 send showMediaWindow Object
instrument.js:108 main set pos
instrument.js:108 main Object
instrument.js:108 main Object
instrument.js:108 on moveMediaWindowToOtherScreen
instrument.js:108 Object
instrument.js:108 output 2
instrument.js:108 pref screen undefined
instrument.js:108 send showMediaWindow Object
instrument.js:108 main set pos
instrument.js:108 main Object
instrument.js:108 main Object
instrument.js:108 on moveMediaWindowToOtherScreen
instrument.js:108 Object
instrument.js:108 output 2
instrument.js:108 pref screen undefined
instrument.js:108 send showMediaWindow Object
instrument.js:108 main set pos
instrument.js:108 main Object
```

mtdvlpr:
@riggles, thanks! Those `Object`s can be expanded to show the contents. I'm interested in the Object next to `showMediaWindow` and the object above `output 2`. Lines 8 and 11 of what you've copy pasted. If you could do the same steps and then show me what those objects contain, I should be able to pinpoint where the issue is.

mtdvlpr:
@riggles, the following comment might be related:
https://github.com/electron/electron/issues/4877#issuecomment-829682127.

How have you configured your screen positions?

mtdvlpr:
https://github.com/electron/electron/issues/34976

sircharlo:
> [electron/electron#34976](https://github.com/electron/electron/issues/34976)

Hm. What's worse. Having rounded corners when using windowed mode, or having to select the target external screen manually?

mtdvlpr:
I would say rounded corners, since you (normally) only have to specify the preferred output monitor once.

sircharlo:
Alright then! In any case I'd call this an edge case, as most KH have Windows computers. If M3 is being used on a Mac, chances are it's for a Zoom meeting, in which case the square corners will work nicely if using the media window in combination with screen sharing (why would anyone do that instead of MP4 sharing, I always wonder?)

mtdvlpr:
Yeah, so we can merge this anyway if we can't fix this issue.

mtdvlpr:
@riggles, thanks! It's weird, because the logs say that he's trying to move the media window to screen 1 (as it should), but it just doesn't move apparently. Afraid I don't have a fix for this, but since you agree it's an edge case and it can be resolved by manually selecting screen 1 in the settings if necessary we'll leave it as is. Thanks so much for all the effort!

Issue #991, opened 2022-12-31
[Enhancement]: Explanation for specific use cases


fatmarker:
### What happened?

JPEGs are being converted to MP4 when the option is enabled.

### To Reproduce

1. Set up a new congregation (English media)
2. Enable convert videos to mp4
3. Generate a weekend meeting media folder

### What did you expect to happen?

Image files should not be converted.

### Version

v22.12.1

### What platform are you seeing the problem on?

MacOS

### Screenshots

<img width="711" alt="Screen Shot 2022-12-31 at 20 35 35" src="https://user-images.githubusercontent.com/4668417/210154064-589f8d4f-e4e3-4a43-b611-76ab9e4fa90b.png">

### Additional context

**EDIT**

A new user (like me) might be confused by the MP4 conversion feature. Provide help tips on this setting for additional explanation.

mtdvlpr:
Hi @fatmarker, this is actually intended behaviour. The use case is for fully remote Zoom meetings where you can use the share mp4 video feature of Zoom to easily share media. By converting images to videos, you can use Zoom's feature for all media.

fatmarker:
I see. Then let's change this to a UX improvement. As this is a very specific use case, an explanation of this setting could be included in a help tip.

mtdvlpr:
Good suggestion! I'll see what I can do and keep you posted :)

mtdvlpr:
Hi @fatmarker, I have a version of this feature for you to test. You can find it [here](https://drive.google.com/drive/folders/1_2CzKieYGaK3LcYrEwzNQgfTpNReOSx9?usp=sharing). I have added this explanation feature to the `convert media to MP4` setting. Please let me know what you think of it!

fatmarker:
Perfect! 👌🏼

PR #1000, opened 2023-01-02
feat: add explanation tooltips to settings

mtdvlpr:
## Proposed Changes

Adds more explanation about certain settings by means of a question mark icon with a tooltip, similar to the locked icon for settings locked by the congregation server settings. It's currently implemented to only show the information icon when the setting is **not** locked. This is intentional, both to keep the UI clean and because extra information for a setting that you can't alter is not really useful.

Currently, I added information for `convert media to mp4` (since that's the setting that led to this feature request) and `hide media win after media finished playing` since that's a really specific use case that is not immediately clear (sign language).

Related to #991


sircharlo (approved):
Love it. Eventually I see other toggles potentially getting some help text as well. Awesome.

Issue #1020, opened 2023-01-04
[Feature]: zoom part mode in presentation mode


mtdvlpr:
Or maybe an on/off in-person toggle in the media controller. 
If off (not in person, Zoom part), then hide the OBS scene picker, set the scene to Zoom scene and always hide the media window after media is done.
If on (in person, not a Zoom part), then show the OBS scene picker. The OBS scene picker would not have the "media" or "Zoom" scenes in the list, to avoid confusion.

_Originally posted by @sircharlo in https://github.com/sircharlo/meeting-media-manager/discussions/987#discussioncomment-4594454_

PR #1022, opened 2023-01-05
feat: add zoom scene

mtdvlpr:
## Proposed Changes

Adds a zoom scene selector to the settings. If filled in, a zoom part toggle button will be present in the presentation mode.

When zoom part is enabled the following things will differ from the default behavior:

- The scene picker will be hidden and the zoom scene will be selected
- The media window will automatically hide after showing media
- After showing media and when pausing a video, the obs scene will return to the zoom scene
- The "hide for zoom only" button for images will be hidden

Related to #1020


sircharlo:
Works awesome based on my tests! Check out my minor tweaks to see what you think

mtdvlpr:
@sircharlo, made the title a little more concise to prevent saying the same thing twice :)

sircharlo:
I see no issues here! I'm sure with feedback, there will be tweaks along the way but you built a solid foundation with this

Chipperchap:
May I Beta test please...?

mtdvlpr:
But of course!
https://drive.google.com/drive/folders/1_2CzKieYGaK3LcYrEwzNQgfTpNReOSx9

Chipperchap:
https://user-images.githubusercontent.com/35364657/210900709-f3a542c7-1769-40bf-8316-a19943a75172.mp4

I don't have much of a graphics thing in my PC so the screenshot video is a bit glitchy but the basics are there. 
I have the sequence wrong in the video - I should show the main hall camera first and immediately after the "Public Talk" but we will get that right eventually. 
Thank you team - this is going to be a wonderful feature - so much easier! (Due to a second wave of Covid we have had over the last few months a few speakers give us hybrid talks.)

Chipperchap:
I don't have much of a graphics card so quality is mediocre but the idea is there. 😁


https://user-images.githubusercontent.com/35364657/210903155-41bc2d88-e08c-4884-9654-9a01ad8a076d.mp4

mtdvlpr:
@Chipperchap, thanks for the video! Three things
1. Should we automatically hide the yeartext when activating Zoom part mode?
2. Should we auto activate media window when leaving Zoom part mode?
3. The blue line appearing under the bottom song is a bug, thanks for making me aware of it!

Chipperchap:
> Should we automatically hide the media window when activating Zoom part mode?
1. It probably makes sense that when the Hybrid button is selected, then the display media mode is automatically stopped. 

> Should we auto activate media window when leaving Zoom part mode?
2. I think that anything that is going to make it easier for less technical is helpful. Instead of multiple clicks to start / stop each. I can just imagine the operator being nervous and forgetting one step and not realizing for a while.

> The blue line appearing under the bottom song is a bug, thanks for making me aware of it!
3. I noticed that but knew you would be looking for things that I wouldn't necessarily notice!

PR #1025, opened 2023-01-06
Fix/several

mtdvlpr:
## Proposed Changes

The following fixes:

- Make video progress bar readonly
- Auto show media window after leaving Zoom part mode

### DIFF
--- a/src/renderer/components/media/MediaItem.vue
+++ b/src/renderer/components/media/MediaItem.vue
@@ -95,7 +95,7 @@
         />
         <v-progress-linear
           v-else
-          v-model="progress"
+          :value="progress"
           absolute
           bottom
           aria-label="Video progress"
--- a/src/renderer/pages/present.vue
+++ b/src/renderer/pages/present.vue
@@ -135,6 +135,9 @@ export default defineComponent({
     showButtons(): boolean {
       return this.shortScenesLength < this.availableWidth
     },
+    mediaWinVisible(): boolean {
+      return this.$store.state.present.mediaScreenVisible
+    },
     showShortButtons(): boolean {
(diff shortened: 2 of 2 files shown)


mtdvlpr:
These are logical. Showing the media win automatically can't hurt and makes sense because the user indicates that the Zoom part is done. Hiding the media window automatically when turning Zoom part mode on is a bit riskier. We assume that they set up the pin and spotlight in Zoom first before activating the Zoom part mode. It's a sensible assumption though, because you need the pin and spotlight for the Zoom scene (which is auto activated) to make sense.

Related to https://github.com/sircharlo/meeting-media-manager/pull/1022#issuecomment-1373289427

Chipperchap:
While the return to normal mode could be a one click option that automatically activates the media display, perhaps the **_entering_** the Hybrid mode could be a double click or a "are you sure?" option. The "are you sure" popup or hover label could contain a warning such as: "Please confirm ZOOM 'Pin video' and 'Spotlight Video' are ready –> Okay”
Having said that, those who use this option / feature would hopefully be aware of the prepare steps involved anyway?

sircharlo:
> While the return to normal mode could be a one click option that automatically activates the media display, perhaps the **_entering_** the Hybrid mode could be a double click or a "are you sure?" option. The "are you sure" popup or hover label could contain a warning such as: "Please confirm ZOOM 'Pin video' and 'Spotlight Video' are ready –> Okay”
> Having said that, those who use this option / feature would hopefully be aware of the prepare steps involved anyway?

I was just going to suggest exactly this.

mtdvlpr:
@sircharlo, what do you think about this?

sircharlo:
I would maybe just say "participants" instead of "window", and

PR #1150, opened 2023-01-11
feat: automate screenshots

mtdvlpr:
@sircharlo,

The result of this automation can be seen at 4b354e9e01e20f927c761b51bd965bbc79d30feb. Will now be executed every time a new version of M³ has been released.

It can be tweaked further, like changing any preference/setting (present mode, meeting day, shuffle music, etc) before taking a screenshot and more images could later be automated as well. But for now, this is a nice start I'd say :)


mtdvlpr:
@sircharlo, I'm not sure this is what you intended. Is it? 
music-playing

sircharlo:
> 

No of course not lol; I was just playing around to see if some sort of highlight could be applied. I just wanted to see the result. 
Something like [this](https://stackoverflow.com/questions/73195502/how-do-apply-border-around-the-element-in-playwright) may be doable, but I don't know if it's worth it.

mtdvlpr:
> > 
> 
> No of course not lol; I was just playing around to see if some sort of highlight could be applied. I just wanted to see the result. Something like [this](https://stackoverflow.com/questions/73195502/how-do-apply-border-around-the-element-in-playwright) may be doable, but I don't know if it's worth it.

If you want to test some things out locally, I suggest running `yarn build-dev` and then `yarn test [folder]` where folder is the folder of which test you want to execute, for example `yarn test home` to only run tests inside `tests/home` or `yarn test shuffle` to only execute tests inside the shuffle.test.ts file. The screenshots will be saved to an img folder at the root.

Issue #1250, opened 2023-01-18
[Feature]: Alternative media numbering for WE-meetings


stepstra:
### Is your feature request related to a problem?

More or less ... ;-)

### Describe the solution you'd like

I'd like the media numbering for weekend meetings to start with 03-01 for the 1st wt song, then 04-xx for the wt-images and finally 05-01 for the 2nd wt song. This way, 01-01 could be reserved for the public talk song and 02-xx for public talk images. This would be consistent with the numbering for the midweek meeting, either.

### Describe alternatives you've considered

I manually change the numbering, sometimes directly in the filename, sometimes within M³. But this complicates things, because when you hit "Update media folders" again, the images and songs will get re-imported with their "original" numbering and now you have duplicates.

mtdvlpr:
@stepstra, interesting! I usually add the song under prefix 00-00 and images as 00-xx, but it would make sense to leave room for the song, since it's always included.

mtdvlpr:
@sircharlo, what do you think about this?

stepstra:
Another possibility, if you don't want to predefine the numbers could be the following: start as it is now with 01-01 for the WT Song. If you then add new media and manually give it the number 01-01, M3 should detect it and ask the user, if the following media should all get +1 automatically, so the song would then be 02-01, the first image 03-01, and so on ...

sircharlo:
Simplicity is key. I like the idea of following the same convention as for the MW meeting.
01-00: intro song
02-XX: custom pics
03-00: WT first song
04-XX: WT pics
05-00: WT final song

mtdvlpr:
@stepstra, enhancement has been merged and will be in the next release :)

PR #1261, opened 2023-01-19
feat: improve we day prefixes

mtdvlpr:
## Proposed Changes

For weekend meetings, increase the prefixes by 2 which would allow for the speaker song to be 01-01 and the images to be 02-XX.

Related to #1250


sircharlo (approved):
Tested, works great!

PR #1309, opened 2023-01-25
chore: new translations

sircharlo:


mtdvlpr:
@sircharlo, we might need some force merging here. Seems the macos tests aren't playing nicely...

PR #1311, opened 2023-01-26
Created a proper MacOS iconset

advenimus:
Properly created a new MacOS icon that not only matches the Apple Design template but makes use of the .iconset to .icns function that should be in place.


mtdvlpr:
@advenimus, awesome!

PR #1383, opened 2023-02-07
feat: add headings to media list

mtdvlpr:
## Proposed Changes

Adds headers to the media list to quickly see what media belongs to which part of the meeting (treasures, apply, living, public talk, watchtower).

Related to #1376


mtdvlpr:
@sircharlo, here you go :)

I chose to store the Watchtower titles in the same location as the media for that date in the form of a file with a .title extension and the Workbook headings as a json file in the cache/lang/mwb folder

Issue #761, opened 2022-11-27
[Feature]: Improve jw.org in M3 feature


mtdvlpr:
_Originally posted by @sircharlo in https://github.com/sircharlo/meeting-media-manager/issues/672#issuecomment-1328268004_
     
List of improvements to implement

- [x] Solve desync actions (log in, change lang) #776
- [x] Scale and zoom preview for accurate representation (#1484)
- [x] Show mouse pointer on media window (#1389)
- [x] Solve links (online library / jw hub) opening new browser tabs #776
- [ ] Mirror input (ie., when typing something in search field, languages field, etc)
- [ ] Mirror text selection
- [ ] Zoom in/out

mtdvlpr:
I also think some might find it more useful if the web view had a sort of visual indication of where the mouse pointer is. Sometimes, the point of these demos is to say "click here, click there", and that might be harder for some to understand if they don't see where the pointer is.

mtdvlpr:
- [ ] Solve desync actions (log in, change lang)
- [ ] scale and zoom preview for accurate representation
- [ ] Show mouse pointer on media window

mtdvlpr:
Bump

mtdvlpr:
@sircharlo, nice work!

e-Feitosa:
When opening the JW.ORG window, I miss the window opening maximized. Here in the tests, a small JW window opens on top of the M3, see below how it opens with the screen extended (image 1). Note: On the second screen (TV/MONITOR/ZOOM) the screen is OK. A suggestion (image 2) (I don't know how much work this would take) would be to duplicate what the mouse does on the main screen to the second screen, in the example, on the PC screen, I selected the text, however, on the projected screen, only captures the movement of the course (red circle)

**IMAGE 1**
<picture>
![image](https://user-images.githubusercontent.com/90984283/219206928-97ebf9d3-c5da-4f0b-8b45-9c0457ec851a.png)
</picture>


**IMAGE 2**
![image](https://user-images.githubusercontent.com/90984283/219207796-d4afc348-630c-475a-8df9-9f5004bc3e0a.png)

mtdvlpr:
@e-Feitosa, we'll add the `mirror selected text` to the todo list. As for maximizing th website controller window, the intended behaviour is this:

If the media window is fullscreen, maximize the window.
If not fullscreen, try to match the media window height and width.

Is this not what's happening?

mtdvlpr:
[...]

PR #1389, opened 2023-02-08
feat: add mouse indicator to jw.org in m3

mtdvlpr:
## Proposed Changes

Adds a red circle (can be anything) to indicate where the mouse is when presenting jw.org.

@sircharlo, ~it seems that Electron v23 breaks the present jw.org feature (I have to investigate further), so I reverted it to v22 here.~

Don't forget to run `yarn compile` when testing locally ;)

Related to #761


sircharlo (commented):
Love it! The mouse pointer works very nicely in my tests. 
I noticed though a bug with the website mode in general: if in website mode, the media controller window can be closed with the X button. Then, all windows close except for the website controller window, which itself cannot be closed without getting a bunch of error messages.
Closing media controller window should be blocked while in website mode, I think.

mtdvlpr:
@sircharlo, I agree!

mtdvlpr:
@sircharlo, that should do the trick :)

sircharlo:
This PR can be merged I think, unless there's something else you wanted to add for now?

Issue #1425, opened 2023-02-12
[Feature]:  Quickly add a video from JW.org


sircharlo:
### Is your feature request related to a problem?

When a video needs to be downloaded from jw.org and added to a meeting's media (for example, Governing Body updates), this needs to be done manually. This process could be simplified greatly. 

### Describe the solution you'd like

When clicking on a day's tile, a new category could be added to the existing "Song", "Custom" and "JWPUB" ones. It could be called something like "JW.org Video". 
This category would automatically list videos from JW.org in the target and fallback languages.
Rather than listing all videos (could be overwhelming), the "latest and featured" video feed could be used, seeing as videos that will be used during upcoming midweek meetings are added to this feed automatically. 

### Describe alternatives you've considered

Downloading videos manually also works, but automating the process would allow for the fallback language logic to be used, for the file name to be consistent with other meeting media, as well as for automatic subtitle retrieval.

### Additional context

Perhaps something similar to this snippet I once wrote could be adapted to populate the video list?

```
      $.getJSON("https://b.jw-cdn.org/apis/mediator/v1/categories/" + lang + "?detailed=1&clientType=www", function(videoTypes) {
        let guids = [];
        for (var videoType of videoTypes.categories) {
          $.getJSON("https://b.jw-cdn.org/apis/mediator/v1/categories/" + lang + "/" + videoType.key + "?detailed=0&clientType=www", function(studioFeatured) {
            try {
              if (studioFeatured.category.media && studioFeatured.category.key !== "FeaturedLibraryLandingUnpubLangs") for (var featuredVideo of studioFeatured.category.media) {
                if (!guids.includes(featuredVideo.guid)) {
                  guids.push(featuredVideo.guid)
                  console.log(featuredVideo)
                  console.log(featuredVideo.files.slice(-1)[0].progressiveDownloadURL)
                  console.log(featuredVideo.images.wss.lg)
                  console.log(featuredVideo.title)
                }
              }
            } catch (err) {
              console.error(err)
            }
          });
        }
      })
```

mtdvlpr:
I like it!

mtdvlpr:
Which categories do you suggest we use? `LatestVideos` and `FeaturedLibraryVideos` or more?

Chipperchap:
Oooo, Yes please!

I see a couple of issues though: 
[...]

PR #1427, opened 2023-02-12
feat: quickly add video from jw.org

mtdvlpr:
## Proposed Changes

Adds a 4th option to the manage screen which lists the latest and featured videos from jw.org for easy download. With subtitles if configured, honoring the configured resolution setting, in both the primary and the fallback language.

Related to #1425


mtdvlpr:
@sircharlo, I've worked out something for you to test. I'm currently only using the hardcoded categories `LatestVideos` and `FeaturedLibraryVideos`.

sircharlo:
> @sircharlo, I've worked out something for you to test. I'm currently only using the hardcoded categories `LatestVideos` and `FeaturedLibraryLanding`. What we could also do is add the `StudioNewsReports` category which contains all Governing Body updates (maybe limit it to the latest 2).

Super nice-looking! I'd add `FeaturedLibraryVideos` as well. Those 3 categories should be enough to cover whatever videos could potentially be discussed at congregation meetings.

---

When a video is selected, could we leave JW.org selected, instead of defaulting back to custom?

image

mtdvlpr:
> When a video is selected, could we leave JW.org selected, instead of defaulting back to custom?

@sircharlo, done :)

Issue #1473, opened 2023-02-15
[Feature]: Auto-start meeting music


sircharlo:
### Is your feature request related to a problem?

It would be nice if we could have the option of having the background music (if enabled) start automatically when the app is opened up.

### Describe the solution you'd like

The music should auto-start only if:
- background music is enabled
- a new background music auto-start toggle is enabled
- it is currently a meeting day
- the current time is between [(meeting start time) - (1 hour)] and [(meeting start time) - (music auto fade delay) - (1 minute)]

### Describe alternatives you've considered

Manually starting the music

mtdvlpr:
@sircharlo, yes! One question. You say minus 1 minute. Is that the maximum 1 minute you can select for auto stopping? Does that mean another requirement is that `musicFadeOutType` needs to be set to `smart`?

sircharlo:
> Or is the 1 minute so that music can at least play for one minute and if auto stop is enabled, we should substract that as well?

Yeah, that was my thinking. No use in auto-starting the music if the meeting starts in 3 seconds...

mtdvlpr:
Great! That's how I understood it :)

PR #1478, opened 2023-02-16
feat: auto start music option

mtdvlpr:
## Proposed Changes

Auto start shuffle music if the following requirements are met:

- Shuffle music is enabled and auto start option is enabled
- It's a meeting day
- The meeting starts in less than 1 hour
- The music would play for at least 1 minute after auto starting

Related to #1473


mtdvlpr:
@sircharlo, the one thing I am not sure about is the placement of this logic. It's currently in the shuffle button itself, meaning the logic is executed every time you land on the home screen or the presentation mode screen. I am picturing a scenario where you manually stop the music at the home screen before the meeting (for some reason) and then open the presentation mode only for the music to start up again.

sircharlo:
> @sircharlo, the one thing I am not sure about is the placement of this logic. It's currently in the shuffle button itself, meaning the logic is executed every time you land on the home screen or the presentation mode screen. I am picturing a scenario where you manually stop the music at the home screen before the meeting (for some reason) and then open the presentation mode only for the music to start up again.

I was thinking that it would get triggered at the same time as the automatic sync (if enabled): on app startup.

mtdvlpr:
@sircharlo, gotchu :)

sircharlo (approved):
Tested, works great!! I'm excited for this one 😄

PR #1484, opened 2023-02-16
feat: improve controller view accuracy

mtdvlpr:
## Proposed Changes

- 
- 
- 

Related to #761


sircharlo:
I think the aspect ratio is fixed now.

mtdvlpr:
@sircharlo, nice work! It looks a bit weird while resizing though. Could you try the `resized` event instead of `resize`?

sircharlo:
> @sircharlo, nice work! It looks a bit weird while resizing though. Could you try the `resized` event instead of `resize`?

I was out! Weird how?

mtdvlpr:
Well, while slowly resizing, the aspect ratio keeps changing which causes the window size to change and even be moved to another part of the screen

mtdvlpr:
I won't be able to show you, or test it myself on Windows until tomorrow (it's time to sleep where I am).

mtdvlpr:
@sircharlo, this is what I mean. I'm only resizing, but the window is also moving in random directions. Also, when maximizing the window, the aspect ratio is not recalculated.

[Screencast from 17-02-23 08:39:47.webm](https://user-images.githubusercontent.com/46671786/219581244-a02223f1-1181-4a6f-a3d0-3fa596019839.webm)

mtdvlpr:
I fixed the maximized size by reverting to the previous way of rendering (100% iframe)

mtdvlpr:
> This API already takes into account the difference between a window's size and its content size.

https://www.electronjs.org/docs/latest/api/browser-window

Are you sure we need to set the aspect ratio like this?

mtdvlpr:
@sircharlo, I think we should remove the custom aspect ratio logic, since the resizing acts weird and it doesn't add that much accuracy. The transform by scale makes it very accurate already.

sircharlo:
I just reallllly dislike this kind of thing:

image

Everything is squished. What if like you said, we change `resize` to `resized`? At least the aspect ratio would be set accurately after the resize is complete?

mtdvlpr:
@sircharlo, I haven't tried it yet. That event is not available on Linux, so for Linux I'd say don't set the aspect ratio on resizing.

sircharlo:
> @sircharlo, I haven't tried it yet. That event is not available on Linux, so for Linux I'd say don't set the aspect ratio on resizing.

Good idea. 16/9 for Linux, and calculated aspect ratio for the rest.

mtdvlpr:
@sircharlo, tested the `resized` event and it works nicely!

sircharlo (approved):
In that case, this thing is ready for deployment :)

PR #1518, opened 2023-02-19
feat: add option to toggle media window always on top

mtdvlpr:
## Proposed Changes

- 
- 
- 

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -419,6 +419,9 @@ if (gotTheLock) {
   ipcMain.on('toggleMediaWindowFocus', () => {
     fadeWindow(win, mediaWin)
   })
+  ipcMain.on('toggleAlwaysOnTop', (_e, val: boolean) => {
+    mediaWin?.setAlwaysOnTop(val, 'screen-saver')
+  })
   ipcMain.on('closeMediaWindow', () => {
     closeMediaWindow()
   })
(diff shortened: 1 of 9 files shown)


sircharlo:
I'm generally uncomfortable with this option, for the simple reason that adding options sometimes leads to "incorrect" or overcomplicated usage scenarios.
Why not simply hide the media window when needed to show Zoom for example? 
The feature itself from this PR works well. I'm just unsure if the complication of adding this option if worth the benefit. 
When would this option ever be enabled int the wild? What if some start using it to show Zoom on top of the media presentation window instead of underneath? What if a popup shows up on top of media while presenting?

mtdvlpr:
I totally agree! The request came from the scenario of showing a small timer application. They want to show the timer application on top of the yeartext. Now that I think about it, setting the Windows background to the yeartext would probably also solve that scenario.

mtdvlpr:
I am with you if you want to label it a nonfix

sircharlo:
> I totally agree! The request came from the scenario of showing a small timer application. They want to show the timer application on top of the yeartext. Now that I think about it, setting the Windows background to the yeartext would probably also solve that scenario.

Yeah, good point. I'd lean towards putting this as a `wontfix` like you said. 
The instructions are very clear here and leave no room for interpretation:

- Display the yeartext before the song, after the prayer, and during meeting when no media is being displayed
- Don't combine the yeartext with other artwork, not even artwork from publications.

PR #1539, opened 2023-02-21
feat: add option to receive beta releases

mtdvlpr:
## Proposed Changes

- 
- 
- 

Related to #1526

### DIFF
--- a/.github/workflows/beta-release.yml
+++ b/.github/workflows/beta-release.yml
@@ -1,0 +1,140 @@
+name: 'Release Beta Version'
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '42 10/12 * * *'
+
+jobs:
+  get-commit:
+    name: Get commit message
+    runs-on: ubuntu-latest
+    outputs:
+      msg: ${{ steps.set-msg.outputs.msg }}
+    steps:
+      - name: 'Checkout repository'
+        uses: 'actions/checkout@v3'
+        with:
+          persist-credentials: false
+      - id: set-msg
+        name: 'Set commit message variable'
+        run: |
+          msg=$(git --no-pager log -1 --pretty=format:"%s")
+          echo "msg=${msg}" >> $GITHUB_OUTPUT
+  bump-version:
+    name: 'Bump version'
+    needs: get-commit
+    if: |
+          !startsWith(needs.get-commit.outputs.msg, 'chore(release)') &&
+          !startsWith(needs.get-commit.outputs.msg, 'ci: version bump')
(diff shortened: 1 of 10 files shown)


sircharlo (dismissed):
Looks good, would it be good to add a note like "don't enable this option on the computer that is used to present media at meetings"?

mtdvlpr:
@sircharlo, sure! I just have one thing I have to test with the automatic workflow. I think running it on a schedule is the way to go, instead of after every commit. Every half an hour? Every hour? What do you think?

sircharlo:
> @sircharlo, sure! I just have one thing I have to test with the automatic workflow. I think running it on a schedule is the way to go, instead of after every commit. Every half an hour? Every hour? What do you think?

Once or twice a day is probably plenty, no? Will it publish a new beta release even when there are no commits since the last run? I'm just envisioning a scenario where a user is on beta-259 😆

mtdvlpr:
> Once or twice a day is probably plenty, no? 

Fine by me!

> Will it publish a new beta release even when there are no commits since the last run?

That's what I'm still working on. There's an issue with checking the last commit message (if commit message is a release then there's nothing new).

mtdvlpr:
@sircharlo, I have now added the workflow to automatically execute at 10am (10:00) and 10pm (22:00) UTC. Also, is the warning note good like this?

mtdvlpr:
> perhaps the beta toggle should be hidden if auto updates are disabled?

I don't think so. When auto updates are disabled, you'll still get a notification when a new release is available with the option to install it right away. If you have auto updates disabled, you could still want to receive the beta releases

mtdvlpr:
You can also execute this beta release workflow manually if you really want to release a new beta version. I also found a GitHub action to remove all old beta releases when a new stable version is released. Shall I add that here?

sircharlo:
> > perhaps the beta toggle should be hidden if auto updates are disabled?
> 
> I don't think so. When auto updates are disabled, you'll still get a notification when a new release is available with the option to install it right away. If you have auto updates disabled, you could still want to receive the beta releases

Right on. I forgot that the user is still notified. My bad. (Does it show that I've never used that option? 😄)

sircharlo:
> You can also execute this beta release workflow manually if you really want to release a new beta version. 

Very nice

> I also found a GitHub action to remove all old beta releases when a new stable version is released. Shall I add that here?

Definitely!

mtdvlpr:
> I also found a GitHub action to remove all old beta releases when a new stable version is released. Shall I add that here?
> > Definitely!

Done :)

Issue #1541, opened 2023-02-21
[Feature]: Auto select first document in JWPUB if only one document available


sircharlo:
### Is your feature request related to a problem?

In a JWPUB file, if only one document is available, the user still needs to select it. 

### Describe the solution you'd like

When number of documents is 1, it should be selected automatically to facilitate the process for the user.

### Describe alternatives you've considered

Selecting the document manually

### Additional context

![image](https://user-images.githubusercontent.com/1275666/220374371-0e9d1a40-61e9-45d6-9262-a38051e20082.png)

PR #1544, opened 2023-02-21
fix: correctly add jwpub media

mtdvlpr:
## Proposed Changes

- 
- 
- 

Related to #1541


mtdvlpr:
@sircharlo, that should fix it. The type got reset after the jwpub dialog was finished and the fix to reset media on type change removed the jwpub media as soon as it was added.

sircharlo:
The import works now!

What happened to the tooltip preview though? It's cut off for some reason.

image

image

image

sircharlo:
Very nice! I like it.

PR #1548, opened 2023-02-21
chore: new translations

sircharlo:


mtdvlpr:
@sircharlo, I fear the only solution is to change all the string through Crowdin. Crowdin will just keep overwriting your direct changes to the translations. You might be able to do it faster by [downloading and uploading](https://crowdin.com/project/meeting-media-manager/translations) the files.

sircharlo:
Yeah, interesting that it does that. No matter. I just used the search and replace feature in crowdin, so I think the next sync will be ok.

PR #1604, opened 2023-03-02
feat: use heading in prefix

mtdvlpr:
## Proposed Changes

File prefix changed from `part-nr` to `heading-part-nr` where heading is either treasures (1), apply (2) or living (3), part is nth part under the heading and nr is nth media of part.

Related to #1593


sircharlo (approved):
Works like a charm!

Issue #1594, opened 2023-02-28
New Scene for Image Presentation


alanfermtz:
### Is your feature request related to a problem?

Right now the applicaton does not let you to have two or more media scene.
Use case:
When we play an image, the zoom is only able to see the image and not the speaker.

### Describe the solution you'd like

Add to the settings an option to have an additional media scene, (only for images), to have a split scene, m3 window and kH camera to still be able to see the speaker even and analyze the image.

### Describe alternatives you've considered

Turning off the zoom image but is not the best because the image is gone

mtdvlpr:
Hi @alanfermtz, thanks for reaching out! Have you considered the "hide for Zoom" feature when presenting images? It will show the speaker for Zoom while still showing the image in the Kingdom Hall. You could toggle what Zoom sees this way.

We could implement your feature, but if above fills your needs, we don't have to add a feature needlessly.

mtdvlpr:
Hi @alanfermtz, thanks for reaching out! We can definitely add this. I'll keep you posted

sircharlo:
Could this not be achieved by having a scene in OBS for "PIP" (or whatever you want to call it) and activating that scene in M3 whenever desired?

mtdvlpr:
The problem is that media is active. While media is active, the scene picker in M3 is ignored until the media is done.

alanfermtz:
> The problem is that media is active. While media is active, the scene picker in M3 is ignored until the media is done. You could in theory use the "hide for Zoom" feature and then switch the scene. That would be a workaround for now.

That is the workaround we are using right now. And its all right really but if we have this, it will awesome. Before using m3 we were using obs, zoom, and library. We have these scenes:
Platform Scene
Media Scene
Image Scene
We change this scenes with a keypad based on the situation. Now with m3 is automatically, but we cant have an image scene.

My solution is only based on assumption of how m3 works:
When you play an image, a video, m3 call OBS socket to change to the media scene. You configure that media scene on the settings.

What if you when image is played, m3 call another scene based on the settings, ex: "Scene for images PIP"
And not enforce this image scene, to not affect anyone, if do not want this

I am really sorry for my bad grammar

e-Feitosa:
> Isso não poderia ser conseguido tendo uma cena no OBS para "PIP" (ou como você quiser chamá-la) e ativando essa cena no M3 sempre que desejar?

_Keeping the simplicity of the M3 in the transmission of images I think is very important._

[...]

PR #1605, opened 2023-03-02
feat: add obs image scene

mtdvlpr:
## Proposed Changes

An optional image OBS scene picker that will be used when presenting images.

Related to #1594


mtdvlpr:
@sircharlo, what do you think of this feature?

sircharlo:
> @sircharlo, what do you think of this feature?

Works really good! I added a quick fix for showing the image scene (if it exists) after hitting the "zoom hide image" button instead of the media scene.

Other than, that, works like a charm. I think it's ready for the big time.

Issue #1671, opened 2023-03-16
[Feature]: Modifying Tab Titles In MMM


athornfam2:
### Is your feature request related to a problem?

N/A

### Describe the solution you'd like

![image](https://user-images.githubusercontent.com/74559383/225672232-873edefd-4c18-4532-a508-74b3ab98ddd3.png)

The items boxed in red are pretty confusing. Would be nice if we could give them better short names for meeting setup and media setup

### Describe alternatives you've considered

N/A

mtdvlpr:
Hi @athornfam2, I agree it can be confusing. The problem is translation. The headings need to fit for every language and some have long translations for short English words. What would you suggest?

sircharlo:
I was actually recently looking into revamping the look and feel of the app. This would also fit into that.

I want to rethink the whole "home screen vs. auxiliary screens" concept of the app into something more akin to [this](https://codepen.io/sircharlo/pen/QWVxgaV).

The user could configure which tab would open by default, which I suspect would be the playback tab for most use cases.

The settings would also be rejigged somewhat, but I'm still thinking about the logic and how to redivide everything more logically for the non-technical end user. 


![image](https://user-images.githubusercontent.com/1275666/225751087-74bc62e1-22a0-456c-9f77-496c79554805.png)
_Logic is hard._

athornfam2:
> Hi @athornfam2, I agree it can be confusing. The problem is translation. The headings need to fit for every language and some have long translations for short English words. What would you suggest?

No real suggestions. We're doing a vendor evaluation of sorts to make sure the application fits our needs as a cong and just saw this sort of stick out as feedback.

mtdvlpr:
@athornfam2, anything else? All feedback is welcome or feature requests are welcome!

mtdvlpr:
@sircharlo, your plans might fit nicely with the [migration](https://github.com/sircharlo/meeting-media-manager/discussions/315#discussioncomment-5181194).

mtdvlpr:
@athornfam2, @sircharlo, what do you think about this?
![Screenshot from 2023-03-17 07-26-21](https://user-images.githubusercontent.com/46671786/225829249-5aea4476-de6f-4b9b-b5d9-225a7c25740f.png)

e-Feitosa:
One option is not to have the ALL tab, and a suggestion is to leave only the name, without including the word configuration (it would include this word on the home screen where there is the gear icon), it will look like this:
System | Server | Media | Meetings

PR #1675, opened 2023-03-17
feat: add icons to setting tabs

mtdvlpr:
## Proposed Changes

Adds icons to the setting tabs to identify them more easily (especially if the initials are the same, e.g. MS and MS)

Screenshot from 2023-03-17 07-26-21

Related to #1671


sircharlo:
Seems like a good fix for now!

sircharlo (approved):

PR #1682, opened 2023-03-18
fix: several fixes

mtdvlpr:
## Proposed Changes

- 
- 
- 


mtdvlpr:
@sircharlo, do you think the "memorial fix" is good? Or are there other edge cases that should throw an error?

PR #1685, opened 2023-03-19
Testing out GUI revamp

sircharlo:
## Proposed Changes

A revamp of the UI may be in order. This PR will collect changes to:
- move to sidebar-based navigation
- revamp the display and categorization of all the settings the app's collected over time

Todo:
- [x] Create rough draft to get an idea of what the nav could look like
- [ ] Fix alignment and look-and feel of elements on congregation selector page
- [ ] Allow selecting a default page on load in settings (either calendar or playback)
- [ ] Go to predefined default page if configured after selecting congregation; if not, go to media playback if media found for today's date, else go to media calendar (currently, selecting a congregation stays on selector page)
- [ ] Allow congregation delete, edit and add from the congregation selector page
- [ ] Make currently selected user (drawer header) a link to the congregation selector page
- [ ] Fix alignment and look-and feel of elements on media calendar page
- [ ] Differentiate tile if that date's folder exists
- [ ] Add toggle for recurring media to settings (disabled by default on new installations)
- [ ] Fix alignment and look-and feel of elements on media playback page (maybe using [this ](https://v2.vuetifyjs.com/en/components/calendars/#usage)as a start-off point)
- [ ] Fix alignment and look-and feel of elements on settings page
- [ ] Make settings page appear to load faster (maybe [this](https://v2.vuetifyjs.com/en/components/treeview/#load-children)?)
- [ ] Re-categorize all settings items and display them in a visually appealing way
- [ ] Potentially allow filtering settings by category (maybe using [chip groups](https://v2.vuetifyjs.com/en/components/chip-groups/#filter-results)?)
- [ ] Create a dynamic "attention required" chip which will appear only when a required field is empty or when a field contains an invalid value and will be assigned to that field
- [ ] If, after opening the app or going to settings manually, some values have the "attention required" chip, activate the filter automatically
- [ ] Allow keyword filter in settings (potential implementation [here](https://v2.vuetifyjs.com/en/components/treeview/#search-and-filter))
- [x] Fix media window (middleware currently breaking media page)
- [ ] Dismiss navigation drawer on click elsewhere; or perhaps make drawer respond to hover?
- [ ] Remove obsolete components and definitions (various buttons and menus will no longer be used)
- [ ] Attempt to use downloaded font for app navigation headers
- [ ] Try to stop the media window flickering when going into settings

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,5 @@
 {
-  "editor.formatOnSave": true,
+  "editor.formatOnSave": false,
   "editor.codeActionsOnSave": {
     "source.fixAll.eslint": true,
     "source.fixAll.markdownlint": true
@@ -24,6 +24,7 @@
     "Bibleand",
     "booknum",
     "cacheable",
+    "Chromecast",
     "CLEARTEXT",
     "clonedeep",
     "codeql",
@@ -135,5 +136,7 @@
     "zipball",
     "русский",
     "українська"
-  ]
+  ],
+  "eslint.codeActionsOnSave.rules": null,
+  "editor.formatOnSaveMode": "modifications"
 }
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
     "@babel/runtime-corejs3": "7.21.0",
     "@electron/asar": "^3.2.3",
     "@fortawesome/fontawesome-svg-core": "^6.3.0",
+    "@fortawesome/free-brands-svg-icons": "^6.3.0",
     "@fortawesome/free-regular-svg-icons": "^6.3.0",
     "@fortawesome/free-solid-svg-icons": "^6.3.0",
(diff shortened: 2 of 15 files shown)


mtdvlpr:
@sircharlo I like your suggested changes! I don't have a clear idea of what your "media playback" and "calendar" pages would be. I suspect media presentation mode and manage media, but where is the "Fetch media" action, in that case? I'm sure you've thought of it and I'm just not seeing it yet.

> Make settings load faster

Might be solved in the migration already, since 1) the watch functions aren't executed on entering and 2) the prefs are immediately correctly inserted.

> Try to stop the media window flickering when going into settings

Should also be fixed in the migration (is because of the watcher functions)

Given this is a big undertaking and would need to be included in the migration as well, I'd suggest only working on this in the migration to prevent a lot of duplicate work. The migration requires a complete rewrite of the code anyway and some default Vuetify styling and alignment changed in v3.

Issue #1730, opened 2023-03-28
[Bug]: Incorrect Watchtower songs are downloaded after Memorial week


mh166:
### What happened?

I went to download the media for our weekend meeting on Sunday, April 16th. And to my big surprise, besides the songs not being in the correct order, we were also suggested a bonus song. 😉 

### To Reproduce

1. Update media for the weekend meeting on 2023-04-16
2. Take a look into the folder
3. Find the following files (in German locale):
```
$ tree
.
├── 03-01 - Lied 96 - Gottes Buch - ein echter Schatz.mp4
├── 04-01 - Absatz 2 - Collage - 1 -  Als Kind hört Jesus seinen Eltern zu - 2 -  Als Jugendlicher hört Jesus zusammen mit seiner Familie aufmerksam zu während in der Synagoge au.jpg
├── 04-02 - Absatz 6 - Collage - 1 -  Eine Schwester schreibt eine Notiz und klebt sie in ihre Bibel - 2 -  Ein Bruder studiert auf seinem Tablet einen Artikel - er markiert Stel.jpg
├── 04-03 - Absatz 14 - Wie können wir feststellen ob wir uns von Gottes Wort formen lassen - Siehe Absatz 14.jpg
├── 04-05 - Lied 95 - Das Licht wird immer heller.mp4
├── 05-01 - Lied 97 - Gottes Wort schenkt Leben.mp4
└── Der Wachtturm Februar 2023 - Wie du mehr vom Bibellesen hast.title
```

### What did you expect to happen?

It should be:
- Song 97 before the Watchtower
- Song 95 after the Watchtower
- Song 96 shouldn't be there at all

### Version

v23.3.2

### What platform are you seeing the problem on?

Windows

sircharlo:
Ah yes, I see what happened. We'll fix it for the next release (to be released much before that date).

mh166:
PS: Same happens the weeks after. Should've tested it right away.

- 2023-04-23: ✅ 144, ✅ 139, ❌ 141
- 2023-04-30: ❌ 097, ✅ 141, ✅ 140, ❌ 95

sircharlo:
@mtdvlpr The problem is with the hardcoded `OFFSET ${2 * weekNr}`. I think we'll have to reqork this to use the "document extract" method, as this allows us to target extracts by document ID. The Multimedia table contains the songs, but they are all assigned to document ID 1, therefore it's unusable.

sircharlo:
[...]

PR #1731, opened 2023-03-29
fix: correctly fetch wt songs for feb 2023 and later

mtdvlpr:
## Proposed Changes

A different approach to get wt songs for wt issues February 2023 and later. I've checked for February and March and also checked if American Sign Language uses KeySymbol 'sjjm'.

Related to #1730

### DIFF
--- a/src/renderer/constants/general.ts
+++ b/src/renderer/constants/general.ts
@@ -45,6 +45,7 @@ export const CHAR_GT = 62
 
 // JW Specific
 export const JAN_2008 = 20080101
+export const FEB_2023 = 20230200
 export const NR_OF_KINGDOM_SONGS = 151
 export const BIBLE_READING_PAR_NR = 12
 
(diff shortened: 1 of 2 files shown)


sircharlo:
Oh wow, I hadn't noticed the change to the databases. Awesome.
I'm suggesting a change in the latest commit here, let me know what you think. 

My idea is, we only want images during the study, so the first query gets only images that are relevant.

The second query gets all videos that are in paragraphs that are also referenced in the DocumentExtract table (long story short, this will eliminate videos from footnotes - see 2023.04.30 for example).

What do you think?

mtdvlpr:
@sircharlo, I think it's good to go now

sircharlo:
Should the output folder be cleared as a migration step before this release? To get rid of the "bad" songs.

sircharlo (approved):

mtdvlpr:
@sircharlo, all tests have run successfully (open their details to see), but there's some [issues with GitHub](https://www.githubstatus.com/). You could forcefully merge it if you want.

PR #1734, opened 2023-03-29
feat: possibility to force critical updates

mtdvlpr:
## Proposed Changes

The question remains, which version do we implement:

When an available update title contains `[critical]`:
1. Immediately quit the app, update and restart
2. Update as soon as the application closes

Or, implement option 1 for `[critical]` and option 2 for `[important]`.

Related to #1732


mtdvlpr:
@sircharlo what do you think?

PR #1897, opened 2023-06-13
chore(beta): fix auto beta release

mtdvlpr:
## Proposed Changes

- 
- 
- 


mtdvlpr:
@sircharlo, the beta release workflow was broken, which caused it to fail immediately. I believe this is the approach we want. `sed '/chore(deps/d'` filters out all commits starting with `chore(deps` and `sed '2,$d'`takes the first line (the latest commit from the filtered list). If the latest commit is not empty and not a release commit, it's either a commit from one of us or a translation commit.

Issue #2003, opened 2023-07-13
[Feature]: Import/Export JW Library Playlists


vangquan:
### Is your feature request related to a problem?

No, the feature request is not related to a problem but rather an enhancement to improve the functionality and interoperability of Meeting Media Manager with JW Library's new Playlist feature.

### Describe the solution you'd like

With the release of JW Library 14, a new feature known as Playlists has been introduced. Each playlist, which can contain images and videos, can be exported as a .jwlplaylist file. I propose we add a feature to import .jwlplaylist files into Meeting Media Manager. Additionally, I suggest implementing a corresponding feature that allows the export of Meeting Media Manager content to .jwlplaylist files.

### Describe alternatives you've considered

As an alternative, we could consider building a separate tool to convert between .jwlplaylist files and Meeting Media Manager files, but integrating this feature directly into Meeting Media Manager would provide a more seamless user experience.

### Additional context

The playlist feature in JW Library 14 seems to be designed with broad utility across different Kingdom Halls/Assembly Halls in mind. Incorporating the ability to import/export .jwlplaylist files into Meeting Media Manager could significantly enhance the flexibility and interoperability of our media management.

mtdvlpr:
Hi @vangquan, I agree! We'll definitely look into this. Any help with figuring out how to read a `.jwlplaylist` is welcome.

sircharlo:
> Hi @vangquan, I agree! We'll definitely look into this. Any help with figuring out how to read a `.jwlplaylist` is welcome.

Nothing complicated; it's essentially a zip file. Contents include a database and any pictures. Videos are referenced in the database so they can be downloaded by API on each client.

mtdvlpr:
> Nothing complicated; it's essentially a zip file. Contents include a database and any pictures. Videos are referenced in the database so they can be downloaded by API on each client.

Oh, so just like a jwpub file, then. Could've known...

mtdvlpr:
@sircharlo, do you also happen to know already how we could export to a playlist file?

sircharlo:
> @sircharlo, do you also happen to know already how we could export to a playlist file?

We would just have to create a database based on the contents of the media list. Doesn't seem to be too complicated to reverse-engineer. I'm just wondering when an export from M³ would be useful...

I can see an import feature being useful, definitely. But export? What scenarios would that be used for?

mtdvlpr:
> I can see an import feature being useful, definitely. But export? What scenarios would that be used for?

[...]

PR #2028, opened 2023-07-16
feat: support jwlplaylist files

mtdvlpr:
## Proposed Changes

- 
- 
- 

Related to #2003


mtdvlpr:
@sircharlo, I've made a first setup for importing `.jwlplaylist` files. I tested it with a JWL video, JWL image, local video and local image and it seems to work correctly. I'll admit, it's not pretty. I'd love your feedback on it.

sircharlo:
Very nice! Basic, but does what it needs to. I added a quick and dirty prefix to make sure the files are listed in the same order as in the jwlplaylist file.

sircharlo:
Quick note: I tested with a playlist from my phone including a local picture taken with my camera in HEIC format. This didn't show, as we don't support HEIC as of yet. However, the same playlist file on JWL works great on Windows. It seems they convert HEIC on the fly or something. 

Should we look into adding HEIC support with something like https://github.com/catdad-experiments/heic-convert or https://github.com/alexcorvi/heic2any?

mtdvlpr:
@sircharlo, we could do that, sure!

PR #2994, opened 2024-10-27
refactor: split electron process into separate files

mtdvlpr:
## Proposed Changes

This PR cleans up the Electron source files, by splitting up the electron process into several files.

- Move the Browser Window logic to a base, main and media file
- Move the updater logic to a separate file
- Move re-usable constants to a separate file
- Move re-usable utility functions to a separate file
- Move the session logic to a separate file
- Move the ipc logic to a separate file

Furthermore, I have implemented some IPC logic in the recommended and secure way (with type safety). `@electron/remote` should not be used as it has many issues and is not secure (see [Electron remote warning](https://github.com/electron/remote?tab=readme-ov-file#electronremote) and [Electron security docs](https://www.electronjs.org/docs/latest/tutorial/security#4-enable-process-sandboxing)).

Refactoring the entire preload script to get rid of `@electron/remote` is too much work, so I plan to gradually refactor each preload function. Once that's done, we can get rid of `@electron/remote` and `sandbox: false`.


sircharlo:
Looks great!! Thanks so much for helping. I can't tell you how much I appreciate.

PR #3002, opened 2024-10-27
Added the jw.org terms of use reference to the Readme

ux-git:
For some people, this point is decisive, so having it right there in the repository is definitely useful


mtdvlpr:
Thanks! I changed the url to automatically open the Terms of Use in the browswr language of the user.

PR #3075, opened 2024-10-31
fix: no refs outside vue

mtdvlpr:
## Proposed Changes

This PR removes all `storeToRefs` from helper and store code. `Refs` are Vue-specific features, like `onMounted` and `computed`, which means that they can only be used inside a Vue context (components and composables).

This PR might break some things, so it has to be tested before merging.


mtdvlpr:
I've tested this and everything seems to be working.

PR #3183, opened 2024-11-07
feat: ability to send announcements to users

mtdvlpr:
## Proposed Changes

Ability to send announcements to users of all versions, using the `announcements.json` file in the master branch of the repo.

## Todo

- [ ] Replace placeholder `warning.svg` with one from the official design


sircharlo:
Looks great! 
Question I just thought of (since I don't have a Mac anymore): does the autoupdater show a notification on macOS when a new version is available? I understood that the update mechanism does nothing on macOS since the app isn't signed.
If that's the case, would it be worth adding a check to see if a new version is available and showing a banner to the user in that case, but only on macOS?
Just a thought.

mtdvlpr:
> Looks great! Question I just thought of (since I don't have a Mac anymore): does the autoupdater show a notification on macOS when a new version is available? I understood that the update mechanism does nothing on macOS since the app isn't signed. If that's the case, would it be worth adding a check to see if a new version is available and showing a banner to the user in that case, but only on macOS? Just a thought.

Done :)

sircharlo:
Loce this!

sircharlo (commented):
🎉

joaochris:
@sircharlo Is this an update notification? 😅

mtdvlpr:
@joaochris, this allows us to send any type of notification to users without the need to update the app. Similar to the Breaking News feature of JW Library.

We've also added an automatic update notification for macos users, because automatic app updates are not supported for macos.

PR #3231, opened 2024-11-08
Update README

ux-git:
## Proposed Changes

Specified that the terminal commands should be executed one after the other, for non-very-technical friends


mtdvlpr:
@ux-git, if you copy paste the lines, don't they automatically run one after the other? If not, we should split them into two separate code blocks

ux-git:
Oh, you're right 👍

Issue #3232, opened 2024-11-08
[Feature]: Do not switch the OBS scene via M³ while media is playing


fschein:
### Is your feature request related to a problem?

In the previous version of M³, when selecting another scene directly from M³, the scene didn’t switch immediately. I used it as "next scene when the video finishes." In the current version, M³ switches the scene immediately.

### Describe the solution you'd like

If I want to switch the scene immediately, I can use OBS. If I want the scene to change after the video finishes, I can select the next scene in M³, just like in the previous version.

### Describe alternatives you've considered

One alternative could be adding an option or checkbox in M³ settings that lets the user decide whether the scene should switch immediately or only after the video ends. This way, users can configure the behavior based on their needs.

mtdvlpr:
Duplicate of #3214

ux-git:
@mtdvlpr The "Switch scene immediately or only after the video ends" setting sounds pretty good though

fschein:
Thank you for addressing this issue. I understand that this behavior was discussed previously, and I appreciate the feedback. However, I would like to suggest that adding an option to control scene switching behavior within M³ could be beneficial for many users. While the current setup is fine for those who prefer immediate scene switching, there are users, including myself, who found the previous behavior (where scenes would switch only after the media finishes) more useful for certain situations. Thank you for considering this enhancement!

sircharlo:
In such a case that we add this setting and it's enabled by the user, I suppose the media scene would no longer be displayed on the user in the popup? The popup would serve as a way to change the scene immediately if no media is playing, or after media is stopped, correct?

mtdvlpr:
Yeah, it wouldn't make sense to show the media scene as a choice when it's always enforced during media.

sircharlo:
> Yeah, it wouldn't make sense to show the media scene as a choice when it's always enforced during media.

My thoughts exactly

fschein:
This issue can be considered low priority. I miss the feature, but it doesn't require effort at the moment.

sircharlo:
Great news, @mtdvlpr added this feature for you!

![image](https://github.com/user-attachments/assets/0a8e0c85-1090-4b2f-9716-0f16135b0e0c)

ux-git:
@sircharlo what a real brotherly love 😊😄

fschein:
I'm speechless. Thank you so much 😊

PR #3284, opened 2024-11-13
feat: add obs scene behavior toggle

mtdvlpr:
## Proposed Changes

Adds support for pre-selecting a new camera scene while media is still playing.

- A "Don't switch scenes during media playback" setting to be toggled.
- When activated, media scenes will be hidden from the OBS popup.
- When activated and an obs scene is selected while media is playing, the scene will be saved and shown as active, but only be sent to obs after media has stopped playing.

Related to #3232


sircharlo:
The feature works great!

I hope you don't mind, I changed the phrasing a bit.

PR #3338, opened 2024-11-15
feat: Automatic numerical sorting of watchedFoldermedia

advenimus:
- Added the functionality of automatically sorting watched media folder items based on numerical file names (ie. 1 - Song 28.mp4, 2 - Photo 1.jpg, etc)
- Reset of order will restore based on this numerical order as well.

CleanShot 2024-11-14 at 19 57 26


mtdvlpr:
hi @advenimus, thanks for your contribution! I have implemented a similar feature in https://github.com/sircharlo/meeting-media-manager/commit/1f566750e67c89ec0ead2c14fd882fe18248ae17, only the sorting does not require a numerical sequence, but simply sorts based on the entire filename. Does this solve your use case?

mtdvlpr:
P.S. What do you think of our [CONTRIBUTING.md](https://github.com/sircharlo/meeting-media-manager/blob/master/CONTRIBUTING.md)? Was it clear? Is it missing something?

mtdvlpr:
> hi @advenimus, thanks for your contribution! I have implemented a similar feature in [1f56675](https://github.com/sircharlo/meeting-media-manager/commit/1f566750e67c89ec0ead2c14fd882fe18248ae17), only the sorting does not require a numerical sequence, but simply sorts based on the entire filename. Does this solve your use case?

This would break when you have 10+ files ("10 file" would be put before "2 file"), but can be solved by using a preceding "0" ("01 file", "02 file").

sircharlo:
> > hi @advenimus, thanks for your contribution! I have implemented a similar feature in [1f56675](https://github.com/sircharlo/meeting-media-manager/commit/1f566750e67c89ec0ead2c14fd882fe18248ae17), only the sorting does not require a numerical sequence, but simply sorts based on the entire filename. Does this solve your use case?
> 
> This would differ from your solution when you have 10+ files ("10 file" would be put before "2 file"), but can be solved by using a preceding "0" ("01 file", "02 file").

I've improved the logic further, using Intl.Collator, so natural sorting will take into consideration numbers like "01", "2", and "003" correctly. (2a4753edb6410fd5ff083f32446e6bd6cde423b7)

advenimus:
> P.S. What do you think of our [CONTRIBUTING.md](https://github.com/sircharlo/meeting-media-manager/blob/master/CONTRIBUTING.md)? Was it clear? Is it missing something?

Was clear yes, only think I think may be needed is if you want a prefix such as fix: or feature: added to request name (wasn't sure about that)

advenimus:
@sircharlo - When could this be added to primary release build? This is the only hold back for us to migrate to new version.

sircharlo:
> @sircharlo - When could this be added to primary release build? This is the only hold back for us to migrate to new version.

I'm planning to release v24.11.4 later today!

sircharlo:
In that case, I'll close this PR, since the changes are already implemented. 😄

PR #3348, opened 2024-11-15
feat: add media from study bible

mtdvlpr:
## Proposed Changes

This PR adds a study bible picker that gets media from the study bible media gallery.

Caveats:

- A lot of study bible media is only available in English, but since it's mostly images, I do want to include it for every language
- When the media items are not available in the media language, they will be shown in English. Images will have the English label and videos will be fetched in the media language (assuming it exists in the media language).

Todos:

- [x] Localize media item information when available
- [x] Support multi-select


mtdvlpr:
@sircharlo, I made a start for a study bible media picker. Maybe you can help me to get it to the finish line. I still to actually download the media once selected and maybe there's a way we can localize the bible books and media calls.

mtdvlpr:
I found 3 scenarios:
- image with direct url
- video with cdn url where we can extract the paramters from
- video with docid/pub and track combo

mtdvlpr:
I've managed to get the localized Bible books! I'm going to log off for now...

mtdvlpr:
@sircharlo, let me know what you think :)

sircharlo:
Wow, you're on fire!!
I'll take a look at this on Monday.

ux-git:
@mtdvlpr not sure if it's already implemented, but having the multi-select available would be a nice boost to the user experience. Instead of opening the window 2 times to add 2 images:

Screenshot 2024-11-16 at 20 47 53

mtdvlpr:
@ux-git, good one! This is currently not implemented yet.

ux-git:
> A lot of study bible media is only available in English, but since it's mostly images, I do want to include it for every language

Meaning the images (not videos) could be downloaded even for unsupported languages? Also, the text descriptions will be hidden in such case?

mtdvlpr:
> > A lot of study bible media is only available in English, but since it's mostly images, I do want to include it for every language
> 
> Meaning the images (not videos) could be downloaded even for unsupported languages? Also, the text descriptions will be hidden in such case?

Yes. Currently both videos and images are shown in English. Videos are downloaded in the media language with the proper label (I'm currently assuming the video exists in the media language). Images are downloaded with the English label.

I figured better a description in English than no description at all.

sircharlo:
This looks great! I added a chapter picker, let me know what you think.

sircharlo:
Added multi-select. @ux-git thoughts?

mtdvlpr:
@sircharlo, awesome! Pushed a small fix. I think it's good to go :)

ux-git:
@sircharlo 

> Added multi-select. @ux-git thoughts?

How can I view this build to generate some thougts?

sircharlo:
> @sircharlo
> 
> > Added multi-select. @ux-git thoughts?
> 
> How can I view this build to generate some thougts?

Here's the build including the latest tweaks; should be out of the oven in about 5 minutes.

sircharlo:
image

---

image

---

image


---

image

ux-git:
@sircharlo looks awesome! The only thing that I can immediately see – could you add some more margin for the checkbox? For it to have some distance from the edges of the image.

sircharlo:
Like this?

image

ux-git:
Exactly 👌

ux-git:
Both work for me. The second one is a bit cleaner and visually more compact, especially when selecting 2, 3 or more images at once.

sircharlo:
LGTM!

PR #3357, opened 2024-11-15
Key Benefits at a Glance - update Readme

ux-git:
Included key benefits, a kind of TL;DR for quick overview of real-word benefits.


sircharlo:
Love it!

PR #3403, opened 2024-11-19
refactor: bump quasar/app-vite and eslint

mtdvlpr:
## Proposed Changes

A major version bump for `@quasar/app-vite` and `eslint`

- Should be a lot better/faster/nicer overall
- Uses the latest technologies (future ready)
- Electron main process is compiled to modern ESM instead of outdated CommonJS
- Better TS typings overall
- Vite 2 > Vite 6


mtdvlpr:
@sircharlo, this should be good to go now. If you could test it as well, this would be a big jump forward dependency wise.

sircharlo:
Looks great from here! Thanks so much for doing this!

PR #3426, opened 2024-11-20
feat: ask for translation help for untranslated languages

mtdvlpr:
## Proposed Changes

This PR adds an announcements banner for media languages that are not yet supported as display language. The banner is not persistent, so only shows once (until closed) per congregation per language).


sircharlo:
This is great stuff!

sircharlo (approved):
LGTM!

Issue #3427, opened 2024-11-20
[Feature]: Ability to add missing media from jwpub files


ux-git:
### Is your feature request related to a problem?

While importing a jwpub file with unlinked videos (e.g. pioneer school media overview file), there are like 20 errors all together. Even though they go away in next few seconds, when video operator should act quickly, those errors blocking the whole app's view.

@sircharlo @mtdvlpr or is there any other workaround to import such tricky files? 

### Describe the solution you'd like

Add "Close all errors" button

### Describe alternatives you've considered

1. Wait, 2. Close one by one

### Confirmations

- [X] I confirm that I have read the documentation website.
- [X] I confirm that I have searched through the existing issues to make sure my feature request is not a duplicate.
- [X] I confirm that I have searched through the existing discussions to make sure my feature request is not a duplicate.

sircharlo:
Do you have a screenshot?

ux-git:
@mtdvlpr not the same error. Different 20-ish media can't be imported. Basically, the issue is the same, but the videos are different.

ux-git:
Exactly

mtdvlpr:
In the previous version, we had an extra step that would indicate which files are missing with the option to provide those files from the local system. Similarly to how JW Library handles it, but in bulk. We could probably re-introduce this.

mtdvlpr:
@sircharlo 

![image](https://github.com/user-attachments/assets/99a8ea25-598c-45ba-8580-1a07c3bcff4e)

mtdvlpr:
@ux-git, in my case, there was some duplication. Each error was shown 6 times. Also, some file identifier info was missing (docid). I've fixed that in eb7deef8401c7a6f81569f7817e0cc7b53ae26ac

sircharlo:
> In the previous version, we had an extra step that would indicate which files are missing with the option to provide those files from the local system. Similarly to how JW Library handles it, but in bulk. We could probably re-introduce this.

We could reintroduce the missing files step with a streamlined approach. 

- A non-persistent banner could be displayed for that day's media whenever a JWPUB file is imported that has missing media. 
- This banner would show a reactive list of missing videos that haven't been added yet. 
- Each item in the list could include a "Browse" button, allowing the user to select the missing video manually. 
- Once selected, the video would be added to the correct spot, and the corresponding item would disappear from the list. 
[...]

PR #3435, opened 2024-11-20
feat(jwpub): display missing media

sircharlo:
## Proposed Changes

-
-
-

Related to #3427


ux-git:
@sircharlo do you need any design input for the missing media placeholders or something else? Or if you want, you can do it like you wish and then share the screenshots, and I'll share my quick feedback and perhaps some visual suggestions if applicable.

sircharlo:
> @sircharlo do you need any design input for the missing media placeholders or something else? Or if you want, you can do it like you wish and then share the screenshots, and I'll share my quick feedback and perhaps some visual suggestions if applicable.

I'll never say no to feedback! Want screenshots or a test build?

ux-git:
For now screenshots would be better because I'm always keep rushing here and there :)

ux-git:
When thinking about intuitiveness, I have the following approach on my mind:

1. All the controls and title are grey highlighting that they're just inactive. Not red as an error or mistake, but rather inactivity.
2. The preview image is pretty much faded out, to emphasize that the media is not actually there there (thats the point that is currently very unclear in the Library, user has to first click to learn that the video is actually unavailable).
3. The explainer text in written just below the image description/name. Reasons: a) we have a lot of space there, b) it's immediately accessible and explains what to do, no need to search for the hover hint (otherwise user must also come up with an idea to hover over the icon first).
4. In the middle of the thumbnail we can place the "Import first" text with an icon to make it even more clear what to do. By clicking on it the import window can open right away. This is to make it 100% obvious, but I do also like the layout without this button on top because it looks way less busy this way (second attachment).

Screenshot 2024-11-20 at 23 40 33

– or –
Screenshot 2024-11-20 at 23 40 53

sircharlo:
I like it! So no more orange banner on top?

1. Done
2. Done
3. Done
4. I like the second option personally.

image

mtdvlpr:
This is very nice! A little off topic tip: when this is ready to be merged, try `Squash and merge` instead of `Create a merge commit`. This will bundle all the commits of the PR into one commit where the commit message is the PR title. This helps in keeping the master branch commit history clean and therefore also helps in creating the changelog for the next version. This also means you can be as brief or unclear as you want when committing to this branch, since the end result will be one clearly described commit anyway.

ux-git:
Looks good! 

> I like it! So no more orange banner on top?

Maybe it would be better to display a temp error message but orange instead of red, to attract the attention for a short time, shortly explain that "_The file was imported successfully. However, some files are included as placeholders because their media is not directly provided in the import file. You can add the media manually if needed_".

This way the attention will be pointed, quickly informed about the issue and all the further instructions are directly on the media items.

ux-git:
Also, another idea about the custom media badge:

We could integrate the star into the label itself, placing it alongside indicators like "Song," "Footnote," and "Paragraph." This would make the badge’s location predictable and consistent, ensuring users always know where to look for this information. The only clear downside I can see is that we can't add a text like "Custom" there because the label will be too long (and with some translations way longer). Normally we use short numbers or paragraph set of 2 or 3 parapgraphs, that's why any word longer than 5 letters is too long, so perhaps replacing the text with a hover tooltip would do the trick. Your thoughts? 

Screenshot 2024-11-21 at 09 04 33

Alternatively, we could remove the stars entirely, not just for the "Other" congregation type but for the entire "Extra Media" section. Since all media in this section is custom, the stars might be redundant.

To conclude, I think moving the star to the label on the left would be the better solution. It provides clarity while beeing consistent and avoids spreading out the indicators across the interface.

ux-git:
Some other options instead of the start icon, since it's relatively easy to mix it up with the footnote icon. Let me know @sircharlo @mtdvlpr what would you prefer. 

Screenshot 2024-11-21 at 09 40 15

ux-git:
Aaand (sorry 😄) giving it a fresh look today I guess the white overlay can be somewhere around 64% opacity, to make the image a bit better visible (especially when there are multiple in the row, to better distinguish them), but still obvious that it's unavailable.

ux-git:
@mtdvlpr agree, not bad

Screenshot 2024-11-21 at 09 51 25

mtdvlpr:
Let's continue the extra media indicator conversation to #3452

sircharlo:
image

With the latest feedback. Any other suggested changes? 😄

mtdvlpr:
What looks better, the red badge or if the badge matches the banner color?

image

ux-git:
@mtdvlpr I find the red one better

PR #3452, opened 2024-11-21
feat: improve extra media indicators

mtdvlpr:
## Proposed Changes

-
-
-


mtdvlpr:
This is how I've currently implemented it:

image

ux-git:
Instead of mixing multiple icons, we can have a separate icon for the added songs. Same visual point, different inner icon. For me, even the standard note icon is fine for custom songs, but this method covers this slight inconsistency. 

Screenshot 2024-11-21 at 13 58 27

Also, I wouldn't enlarge the icon to keep all of them the same size, which is especially important for weekend meetings where users can easily see all of the icons on one page. I made the icons slightly bigger within the container to if you find the icon a bit too small.

ux-git:
Looks nice. 

@sircharlo what do you think? Does it even make sense to mark the imported song with a different icon?

mtdvlpr:
There's currently also a monitored folder indicator. Should we move that to the left badge as well for consistency?

image

ux-git:
Sounds good! Added the outline icon for the double consistency 😆 

Screenshot 2024-11-21 at 15 05 10

mtdvlpr:
@ux-git, same issue with the monitored indicator. If a song is extracted from a watched jwl playlist file, which badge do we show? I feel like we're over-engineering this a bit. Should we just either show the extra media / monitored icon OR the song icon when both apply?

sircharlo:
> Instead of mixing multiple icons, we can have a separate icon for the added songs. Same visual point, different inner icon. For me, even the standard note icon is fine for custom songs, but this method covers this slight inconsistency.
> 
> Screenshot 2024-11-21 at 13 58 27
> Also, I wouldn't enlarge the icon to keep all of them the same size, which is especially important for weekend meetings where users can easily see all of the icons on one page. I made the icons slightly bigger within the container to if you find the icon a bit too small.

I love this idea.

sircharlo:
> Looks nice.
> 
> @sircharlo what do you think? Does it even make sense to mark the imported song with a different icon?

Yeah, looks great. I personally feel it's very useful to the end user to indicate what is manually / watcher imported vs. automatic.

sircharlo:
What about having a left and right area in the folder icon? Left could show if folder is watched or not, right could show song. Just a thought.

That way, we'd have a few possibilities:
- Watched folder, normal media
- Watched folder, song
- Manually added, normal media
- Manually added, song

mtdvlpr:
Okay, so we have the following options:

Type | Icon
---|---
Song | music note
Extra media | user inside folder
Watched media | start inside folder
Extra song | song inside folder
Watched song | ?

mtdvlpr:
For the watched folder, I think an eye icon would fit best.

ux-git:
Do we even need to show the watched folder media? What is a user case where having the custom media indicator won't be enough?

ux-git:
Maybe simply either custom media OR watched folder media would work? 

Then:
• normal icons for labels of synced media
• watched folder for all media in the watched folder
• custom media folder for all media added manually

 Screenshot 2024-11-21 at 16 19 09

ux-git:
@mtdvlpr would those two icons cover all the main categories/cases?

mtdvlpr:
image

ux-git:
Hmm, the two folder icons are easy to confuse. What do you think about using this Add media icon? Then there will be some logical connection:

Screenshot 2024-11-21 at 16 47 16

sircharlo:
> Hmm, the two folder icons are easy to confuse. What do you think about using this Add media icon? Then there will be some logical connection:

You mean use the Add Media icon for non-watched added media?

ux-git:
> You mean use the Add Media icon for non-watched added media?

Yes

sircharlo:
Like this?

mtdvlpr:
How should I get the correct svg code for the icons? I've been right clicking them and then copy as svg. Is there a better way?

ux-git:
Depending on the workflow: either click on the parent container, and then copy as SVG, or select multiple and click + on the right side Export panel, or use a plugins, I'm pretty sure there's one that shows the code.

ux-git:
> Like this?

Looks better to me. Your thoughts guys?

sircharlo:
> How should I get the correct svg code for the icons? I've been right clicking them and then copy as svg. Is there a better way?

Click on icon in UI Kit > Export > SVG
Then, to show up correctly, the `evenodd` has to be removed (if present) using a tool like [this one](https://codepen.io/herrstrietzel/pen/jOoGrxO?editors=1010).

sircharlo:
> > Like this?
> 
> Looks better to me. Your thoughts guys?

Sounds good!
Are we happy with the repeat indicator being on the right?

ux-git:
> Are we happy with the repeat indicator being on the right?

I think so. Playback indicator near the playback controls. Would it make sense to make it clickable to turn it off? Or is it tricky considering that it must be turned only through the More menu?

sircharlo:
> > Are we happy with the repeat indicator being on the right?
> 
> I think so. Playback indicator near the playback controls. Would it make sense to make it clickable to turn it off? Or is it tricky considering that it must be turned only through the More menu?

Again, I'm torn 😃 
Right now, the indicator is only visible when loop mode is on.

**Option 1:** **Clickable Toggle**  
- **Pros:** Quickly turns loop mode off.  
- **Cons:** Users may disable it accidentally and be unsure how to turn it back on.  

**Option 2:** **Context Menu Option**  
- **Pros:** Maintains consistency with user actions.  
- **Cons:** Slower to disable loop mode when needed.

sircharlo:
(I'm leaning towards option 2 personally.)

ux-git:
> I'm leaning towards option 2 personally.

Me too. If user 1. found this option, 2. and he needs it – then they're advanced enough users.

sircharlo:
LGTM!

Issue #3522, opened 2024-11-24
[Feature]: Nicer scrollbars in the option menu 


dagandrew:
### Is your feature request related to a problem?

Sometimes when the window is small or the title of media is long, the ugly Windows scrollbars appear 
![image](https://github.com/user-attachments/assets/d671b746-f188-47ab-88b8-63982328eed4)

### Describe the solution you'd like

Definitely get rid of the horizontal scrollbar and make the vertical smaller

### Describe alternatives you've considered

Or you could also make the option menu item descriptions shorter, so everything will fit without scrollbars

Or make the menu break out past the window edges, but I think it's difficult as it said here https://www.reddit.com/r/electronjs/comments/m8sadn/is_it_possible_to_break_out_past_the_window_edges/

### Confirmations

- [X] I confirm that I have read the documentation website.
- [X] I confirm that I have searched through the existing issues to make sure my feature request is not a duplicate.
- [X] I confirm that I have searched through the existing discussions to make sure my feature request is not a duplicate.

sircharlo:
I'm in the stages of replacing the hacky custom-scrollbar solution in a way that should fix this issue as well.

sircharlo:
> @sircharlo since this import menu list will gradually become bigger with new features added, could you set no fixed height for this window? To make the card flexible and fit all the content with no need to scroll – everything at a glance. I guess the height of even smaller displays should allow to do this.

You mean have the menu pop out of the main window? We can't do that unfortunately...

PR #3525, opened 2024-11-24
refactor: implement custom scrolling instead of q-scroll-area

sircharlo:
To do:

- [x] Remove all `<q-scroll-area>` tags
- [x] Test dark mode
- [x] Fix modal height so that children can be dynamic height
- [x] Make rows in modals that overflow do so without max-height value

Related to #3522


mtdvlpr:
I've refactored the remaining scrollable content. I think this PR is good to go now :)

sircharlo:
I also think we're good to go 😄

Issue #3481, opened 2024-11-21
[Feature]: Add introduction, index and appendixes to the Study Bible import


ux-git:
### Is your feature request related to a problem?

It would be useful for schools and field service meetings.

### Describe the solution you'd like

Perhaps the tabs can be added, just like in the JWL Bible? Introduction, Books, Index, Appendix A, Appendix B. There are some printed versions of the media which are also often useful.

By following the known pattern we will make it easier to import this additional media.

### Describe alternatives you've considered

Open JW Library instead, or open the jw.org / wol.jw.org

### Additional context

The books layout can be optimized to fit more books and use the window size better by moving the book names inside of the image. And perhaps making the gaps between books smaller. Looks pretty good:

<img width="1278" alt="Screenshot 2024-11-21 at 23 53 02" src="https://github.com/user-attachments/assets/651ea9b8-403c-4328-8af9-211a53b16f5d">

It won't really help with media on other tabs, since there are more desciptive text, but at least for the default tab that will be used the most often it would make sense.

### Confirmations

- [X] I confirm that I have read the documentation website.
- [X] I confirm that I have searched through the existing issues to make sure my feature request is not a duplicate.
- [X] I confirm that I have searched through the existing discussions to make sure my feature request is not a duplicate.

sircharlo:
From what I understand, it seems like this is definitely possible, but we'd have to change the underlying method we use to get media. 

@mtdvlpr Do you think this is worth it? We'd have to go from JSON parsing to JWPUB processing (`getDbFromJWPUB({ pub: 'nwtsty' })`).

sircharlo:
> > @mtdvlpr Do you think this is worth it? We'd have to go from JSON parsing to JWPUB processing (`getDbFromJWPUB({ pub: 'nwtsty' })`).
> 
> It would be nice to have the appendices as well! I tried finding the media gallery refs in the nwtsty jwpub, but couldn't find it. That's why I went for the json approach.

They're all there in the JWPUB , from what I can see...

mtdvlpr:
I'm sure I just wasn't looking in the right place.

mtdvlpr:
I see them now😄. I guess I was either looking at the regular bible or the Dutch study bible before...

sircharlo:
Yeah, we can add it to the roadmap. It'll be a refactor but it's not crucial in any way. Meeting media is the focus of this app :)

sircharlo:
Will be released in the next version!

[...]

PR #3546, opened 2024-11-26
refactor: use JWPUB instead of JSON for study bible lookups

sircharlo:
## Proposed Changes

-
-
-

Related to #3481


mtdvlpr:
Nice work so far! One thing I'm missing (maybe you already thought of it) is the English media when it's not available in the current media language. Do you (have a) plan to support that?

sircharlo:
> Nice work so far! One thing I'm missing (maybe you already thought of it) is the English media when it's not available in the current media language. Do you (have a) plan to support that?

Yeah, I wanted to get everything working in English first as a base, then figure out the rest.

sircharlo:
I think this is in good shape!
Thoughts?

mtdvlpr (approved):
Awesome! It would have taken me a lot longer to figure out the logic... 

I've added the fallback to English media, so that all study media is shown for every language.

sircharlo:
Perfect! I'll merge it. 
Thanks!

Issue #3615, opened 2024-12-02
[Feature]: Select different path for media cache directory


adrianodaddiego:
### Is your feature request related to a problem?

It is currently not possible to specify a custom directory for the media cache. This creates a problem for users who want to store data on a secondary drive, for example, to optimize for space or performance.

### Describe the solution you'd like

Users should be able to define a custom path for the media cache during the initial setup process. This would allow more flexibility for managing storage locations right from the start.

### Describe alternatives you've considered

If there is a workaround is appreciated.  The "extra" folder feature, while useful, does not fully address this need.

### Confirmations

- [X] I confirm that I have read the documentation website.
- [X] I confirm that I have searched through the existing issues to make sure my feature request is not a duplicate.
- [X] I confirm that I have searched through the existing discussions to make sure my feature request is not a duplicate.

sircharlo:
Great idea! @mtdvlpr has implemented this and it is now available in the latest version, v24.12.0!

PR #3622, opened 2024-12-04
feat: add support for custom cache folder

mtdvlpr:
## Proposed Changes

Adds an advanced setting to set a custom cache folder. 

- The cache folder would be used for the `Publications` and `Additional Media` cache. 
- The cache won't be moved from and to a new cache folder, since the current media might still reference the old cache folder.
- Both the default and custom cache folders can be cleared from settings.
- The cache folder setting is present in the setup wizard so it can be set before the first fetching starts.

Related to #3615


mtdvlpr:
@sircharlo, what do you think?

P.S. I think we're ready for v24.12.0 :)

sircharlo:
Looks great! Thank so much for implementing this!

sircharlo (approved):

Issue #3651, opened 2024-12-05
Elapsed and remaining time of a video


joaochris:
### Describe the solution you'd like

Could you return the count of elapsed and remaining time for a video just like it was in the previous version?

![IMG_20241205_202900](https://github.com/user-attachments/assets/7bd3a7e1-7cf3-4f90-b1f7-bb1a29c88f2a)

### Confirmations

- [X] I confirm that I have read the documentation website.
- [X] I confirm that I have searched through the existing issues to make sure my feature request is not a duplicate.
- [X] I confirm that I have searched through the existing discussions to make sure my feature request is not a duplicate.

sircharlo:
Being worked on in https://github.com/sircharlo/meeting-media-manager/pull/3661.

Issue #3593, opened 2024-12-01
[Feature]: Ability to precisily cut longer videos 


dagandrew:
### Is your feature request related to a problem?

I'm frustrated that you can't cut down to the second the video when it's like 30 minutes, it allows only 7 second increments, and if the video is 1 hour it will be like 14 seconds.

### Describe the solution you'd like

it would be nice if you could click on the time above the slider and turn it into input therefore edit it manually there
![image](https://github.com/user-attachments/assets/889361cd-c3c2-455c-b586-a02cd1e7ff12)

### Describe alternatives you've considered

if you want to use only a slider to cut the video, you could also increase the width of the video cutting frame (so it would be alsmot the size of the main window), so the increments could be smaller, but there would be 2-4 seconds if the video is long enough (not perfect of course, but bearable)

### Confirmations

- [X] I confirm that I have read the documentation website.
- [X] I confirm that I have searched through the existing issues to make sure my feature request is not a duplicate.
- [X] I confirm that I have searched through the existing discussions to make sure my feature request is not a duplicate.

sircharlo:
> Alternatively, the start-end time on the sides of the slider can be converted into input fields to allow manual input that is accessible and relatively intuitive.

I think I like that option!

PR #3661, opened 2024-12-06
feat: allow time entry for custom durations; show time elapsed and remaining during video playback

sircharlo:
## To do

- [x] Move customDurations to media props
- [x] Properly set/emit updates to custom duration props
- [x] Fix media player page to respect custom durations

Related to #3593 & #3651


sircharlo:
LGTM!

PR #3944, opened 2025-01-16
chore(readme): mention Bible import features

ux-git:
Included the 1. Audio import, 2. Study Bible import, and 3. Moved S-34 to a separate point for better visibility, since it's once of the key features for users.


mtdvlpr:
Awesome!

PR #4076, opened 2025-02-22
feat: auto Memorial background

mtdvlpr:
## Proposed Changes

This PR attempts to automate the Memorial background with a Memorial date setting.

The background is set in the following order: custom background -> Memorial date background -> yeartext.

Since this would be only the second time that we have a Memorial background, the method of fetching might not be accurate for following years. That's why the setting has been marked as a beta feature.

Related to #4075


sircharlo:
Hey there, sorry, I've been extremely busy these past few weeks! 
Things look like they should stabilize sometime before April 😆 

Thanks for all your help in the meantime! @mtdvlpr

mtdvlpr:
> Hey there, sorry, I've been extremely busy these past few weeks! Things look like they should stabilize sometime before April 😆
> 
> Thanks for all your help in the meantime! @mtdvlpr

No worries! We take turns being busy quite well :)

I'll just go ahead and merge this so it can be tested in beta and subsequently released in time for the Memorial. We can always tweak things afterwards, especially since I've marked it a beta feature.

sircharlo:
Would it make sense to store memorial dates as an array in a JSON file in the repo, like announcements? This would allow for automatic updates as needed, since the dates are the same worldwide.  

If the JSON file is available, the next upcoming memorial date could automatically be set as the memorial date if the currently configured one is empty or over a week old. This would eliminate the need for manual configuration in almost all cases.  

What do you think?

mtdvlpr:
@sircharlo, we could definitely do that! I went with the current approach because it places the maintenance on the user instead of us as maintainers. But your suggestion would improve the user experience greatly.

PR #4705, opened 2025-08-14
chores(readme): Add special events mode and Zoom integration features

ux-git:
Added special events mode and Zoom integration


sircharlo:
Thanks for this!

PR #5287, opened 2025-11-04
feat: new meeting timer feature

sircharlo:
This branch has a new meeting timer feature.

WIP!

---

#### To do

- [x] Prevent having multiple listeners for time/talks update (ex: duraing app refresh); remove existing listeneners to only listen to one, current provider and avoid false/stale data
- [x] Add count-up/count-down setting to display either time as count-up (stopwatch style) or countdown (showing time left)
- [x] Enable gentle non-disruptive lateness indicator (visual indication to speaker if overtime)
- [x] Add configurability for lateness display (change color of bg and/or text when overtime, and/or display the amount of time speaker by which spealer is over instead of the total time used for that part, if in count-up mode)
- [x] PDF report option and/or display time over instead of total time.
- [x] Add ability to set and adjust custom times and parts for non-meeting days
- [x] "Ahead/behind" tracker
- [x] Add setting to see 'current' time/clock display on button
- [x] Add part values for MW meetings:
- [x] Treasures: 10 minutes
- [x] Gems: 10 minutes
- [x] Bible reading: 4 minutes
- [x] Variable AYFM number of parts and timing, total of 15 minutes (timing of all parts + number of parts)
- [x] Variable LAC non-CBS number of parts and timing, total of 15 minutes
- [x] CBS: 30 minutes, or adaptive
- [x] WT: 60 minutes, or adaptive
- [x] Show adaptive button (CBS/WT) in a clear fashion 
- [x] Allow changing adaptive end time easily
- [x] If CO week, WT = 30 minutes, or adaptive, and add CO final talk
- [x] If CO week, CBS = replaced with CO service talk

### DIFF
--- a/package.json
+++ b/package.json
@@ -134,7 +134,7 @@
     "typescript": "~5.9.3",
     "vite": "^8.0.3",
     "vitepress": "^2.0.0-alpha.17",
-    "vitest": "^4.1.1",
+    "vitest": "^4.1.2",
     "vue-eslint-parser": "^10.4.0",
     "vue-tsc": "^3.2.6",
     "webpack-iconfont-plugin-nodejs": "^1.0.36"
--- a/yarn.lock
+++ b/yarn.lock
@@ -11140,7 +11140,7 @@ __metadata:
     upath: "npm:^2.0.1"
     vite: "npm:^8.0.3"
     vitepress: "npm:^2.0.0-alpha.17"
-    vitest: "npm:^4.1.1"
+    vitest: "npm:^4.1.2"
     vue: "npm:^3.5.31"
     vue-eslint-parser: "npm:^10.4.0"
     vue-i18n: "npm:^11.3.0"
@@ -15926,7 +15926,11 @@ __metadata:
   languageName: node
   linkType: hard
 
+<<<<<<< feat/horloge
 "vitest@npm:^4.1.1":
+=======
+"vitest@npm:^4.1.2":
+>>>>>>> master
   version: 4.1.2
   resolution: "vitest@npm:4.1.2"
   dependencies:


alanfermtz:
Is this a timer? like onlyt?
If it is I have a couple of requests if possible:

1. At the end, deliver a pdf with the report like onlyt does.
2. Five minute countdown before the meeting.
3. Adjust custom time, maybe with a template for a convention.
4. Will be nice if you can make the countdown to register if we are behind or ahead of time.
- Ex. for 4: If the Opening Comments are for 1 minute, and the chairman elapsed 1:30, will we have maybe like a little note in the corner that we have +30s on red for the meeting , this "timer" will be a sum of all this gain or loss time.
Image:
223577791-98bf7503-7528-471c-a452-d5593725c8ce

kentestaccount:
> > 1. Main purpose will be shared with others, we currently have OnlyT, and that report help us to see if we need improvement, mainly used by the Auxiliary Counselor. But I think it complies with both purposes really (share and archive).
> > 2. Yes, we have another monitor that serves like clock, timer and countdown that is in front of the speaker. (onlyt)
> > 3. What I am thinking is to be able to not only register regular meeting but also, using it for a convention, not to present the timer for the speaker, but for audio/video department and have a track. So we can customize to fit the times for the speakers (We currently use jmultimedia for this purpose so no big deal if not implemented really)
> > 4. jmultimedia app does it if you need ideas. (https://jwmultimedia.org/descargas/)
> 
> 1. Added to to-do
> 2. Done!
> 3. Added to to-do
> 4. Added to to-do

I’m particularly interested in how suggestion no. 4 could be implemented. In OnlyT, when earlier parts go overtime, the adjustment is usually applied to the Living as Christians section. This works, but it often results in large cuts concentrated at the end.  

It might be more balanced if the adjustments were distributed progressively across all parts of the meeting. For example, if the introduction runs 30 seconds overtime, that extra time could be automatically spread across the following parts rather than deducted in full from the later sections. This would keep the schedule more consistent and prevent sudden, significant reductions at the end.  

This idea seems closely related to [OnlyT’s Issue #540](https://github.com/AntonyCorbett/OnlyT/issues/540), where users requested dynamic time adjustment displays in adaptive mode. If Meeting Media Manager could not only adjust but also distribute overtime progressively, it would make the adaptive timing feature even more effective.

kentestaccount:
Regarding timer screens, the closest direction I’m aware of is in CO-1 Convention Organization Guidelines, Chap. 3, par. 59:

> If the branch office will not be providing a design for the stage, the installation overseer(s) will design it. The design, including decorations, should be simple and tasteful, not distracting in appearance. One or more large-faced clocks that are easily readable by the speaker should be placed on the stage. Do not provide a countdown timer or an electronic signal to the speaker that he is overtime.

Our LBD representatives interpret this as meaning timer screens are not permitted, only analog clocks.  Of course, conventions are different from congregation meetings.

That said, in our local congregation we previously struggled with overtime meetings. Several remedies were tried, but the most effective was using OnlyT. As per direction, approval from the Maintenance Trainer was required before installing anything in the Kingdom Hall. We were told that as long as the digital display replaced the old analog clock and the wiring was up to code, it was acceptable. Perhaps this is a local arrangement, but it shows there can be some flexibility.  

For that reason, I think it’s worthwhile to keep the timer feature in the application. Even if external live screens aren’t allowed everywhere, the adaptive timing logic and display on the A/V desk could still be valuable. That way, congregations that do have approval can benefit, while others can still use the reporting/export functions without conflict.

sircharlo:
> Regarding timer screens, the closest direction I’m aware of is in CO-1 Convention Organization Guidelines, Chap. 3, par. 59:
> 
> > If the branch office will not be providing a design for the stage, the installation overseer(s) will design it. The design, including decorations, should be simple and tasteful, not distracting in appearance. One or more large-faced clocks that are easily readable by the speaker should be placed on the stage. Do not provide a countdown timer or an electronic signal to the speaker that he is overtime.
> 
> Our LBD representatives interpret this as meaning timer screens are not permitted, only analog clocks. Of course, conventions are different from congregation meetings.
> 
> That said, in our local congregation we previously struggled with overtime meetings. Several remedies were tried, but the most effective was using OnlyT. As per direction, approval from the Maintenance Trainer was required before installing anything in the Kingdom Hall. We were told that as long as the digital display replaced the old analog clock and the wiring was up to code, it was acceptable. Perhaps this is a local arrangement, but it shows there can be some flexibility.
> 
> For that reason, I think it’s worthwhile to keep the timer feature in the application. Even if external live screens aren’t allowed everywhere, the adaptive timing logic and display on the A/V desk could still be valuable. That way, congregations that do have approval can benefit, while others can still use the reporting/export functions without conflict.

Thank you for your well thought-out comments; we'll take them into consideration.

ux-git:
Hey @sircharlo, [done](https://github.com/sircharlo/meeting-media-manager/pull/6870)!

sircharlo:
> Hey @sircharlo, [done](https://github.com/sircharlo/meeting-media-manager/pull/6870)!

Thanks! That was fast!

PR #6692, opened 2026-03-11
Zoom automation

sircharlo:
Allow for near complete Zoom automation

### To-Do

- [x] Randomize port number on Python process start, make sure to only use a port if available and not susceptible to be used by another process;  to use this port (instead of `:5000`)
- [ ] Fix menu logical display
- [ ] Fix Zoom icon


src/stores/current-state.ts
```diff
@@ -217,8 +231,59 @@ export const useCurrentStateStore = defineStore('current-state', {
...
+      }
+
+      zoomHelperSyncInProgress = true;
+      try {
+        const pythonInstalled = await isZoomPythonInstalled();
+        if (!pythonInstalled) {
+          const { createTemporaryNotification } =
+            await import('src/helpers/notifications');
+          createTemporaryNotification({
+            caption: 'Please install Python and restart the app.',
```
sircharlo: remove hardcoded string here

PR #6730, opened 2026-03-14
fix: improve congregation lookup language matching for Chinese variants

DarkCodeYG:
## Summary
- Improves language code matching in the congregation lookup dialog for Chinese language variants (CHS, CHT, etc.)
- Previously, selecting a Chinese congregation could match the wrong language variant

## Test plan
- [ ] Search for a congregation using Chinese Simplified (CHS)
- [ ] Verify the correct language variant is selected

🤖 Generated with [Claude Code](https://claude.com/claude-code)


sircharlo:
Thanks for this!

PR #6736, opened 2026-03-15
feat: add dynamic yeartext font loading based on writing script

DarkCodeYG:
## Summary

Load the correct yeartext font per writing script, matching how JW.org handles it.

Noto fonts come from jsDelivr, WT/JW fonts from JW CDN — both cached locally after first download so they work offline. Language-specific overrides (e.g. Traditional Chinese, Karen) are also covered.

No extra bundled fonts, no size increase.

## How it works

MainLayout sends the current script + language to the media player via BroadcastChannel. The media player downloads the needed font (if not already cached) and applies it to `#yeartext`.

Font download and caching reuses the existing `setElementFont()` / `getLocalFontPath()` path — same as JW-Icons and Wt-ClearText-Bold.

## Font sources

- **Noto fonts** (Arabic, Armenian, Bengali, etc.) → jsDelivr CDN, cached locally
- **WT/JW/Manna fonts** (WTTextNew, WTSetthaSpecial, etc.) → JW CDN, URLs discovered from CSS `@font-face`
- **Wt-ClearText-Bold, Wt-BaeumMyungjo** → JW CDN, hardcoded fallback (already in codebase)

## What changed

| File | Change |
|------|--------|
| `src/types/general.d.ts` | Extend `FontName` with Noto + WT font names |
| `src/stores/jw.ts` | jsDelivr URLs in `fontUrls`, `updateYeartextFontUrls()` for WT font discovery |
| `src/helpers/fonts.ts` | `YEARTEXT_FONTS` mapping, language overrides, `loadYeartextFont()` |
| `src/layouts/MainLayout.vue` | Broadcast script + language via BroadcastChannel |
| `src/pages/MediaPlayerPage.vue` | Receive script + lang, load font, apply via `:style` |

## Test plan

- [ ] Korean yeartext shows Wt-BaeumMyungjo
- [ ] Chinese yeartext shows NotoSansSC
- [ ] Roman yeartext unchanged (Wt-ClearText-Bold)
- [ ] Fonts work offline after first download


DarkCodeYG:
Thanks for the guidance!

1. **Bundle size** — Switched to jsDelivr CDN via `@font-face` in `src/css/yeartext-fonts.css`. No packages bundled; CJK uses static weight, others use variable format with `unicode-range`.

2. **Language overrides** — Handled programmatically: `updateYeartextFontUrls()` parses JW.org CSS for `.ms-{SCRIPT}.ml-{LANG}` rules and builds a `yeartextFontOverrides` map. No hardcoded overrides.

PR #7921, opened 2026-06-12
[codex] replace jitsi robotjs with robotjs

sircharlo:
## Summary

- Replace `@jitsi/robotjs` with upstream `robotjs@0.7.1` in the preload dependency path.
- Add a GitHub Actions native-module matrix that builds `robotjs.node` for macOS arm64, macOS Intel, Windows x64, and Windows ia32.
- Add an electron-builder `beforePack` hook that copies the prepared native artifact into `node_modules/robotjs/build/Release/robotjs.node`, which is the path upstream `robotjs` hard-requires.

## Notes

Upstream `robotjs` loads `./build/Release/robotjs.node` directly. The workflow therefore creates a macOS universal binary with `lipo`, and for Windows lets the `beforePack` hook swap the x64 or ia32 native module for each packaging arch.

Linux remains on the normal local build path because the current workflow already installs the X11/png development libraries required for RobotJS.

## Validation

- `yarn lint`
- `yarn test:unit`
- Parsed `.github/workflows/build.yml` locally with the installed `yaml` package

This PR is intentionally a draft so the full GitHub-hosted runner matrix can prove whether upstream `robotjs` is viable for M3 releases.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -87,15 +87,23 @@ jobs:
         include:
           - artifact: robotjs-darwin-arm64
             arch: arm64
+            node-architecture: arm64
+            node-version: 24
             os: macos-latest
           - artifact: robotjs-darwin-x64
             arch: x64
+            node-architecture: x64
+            node-version: 24
             os: macos-15-intel
           - artifact: robotjs-win32-ia32
             arch: ia32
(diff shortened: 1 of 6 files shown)


sircharlo:
Tested on Windows x64 - OK

PR #8658, opened 2026-08-08
fix: make the downloaded FFmpeg executable on macOS and Linux

DarkCodeYG:
## Proposed Changes

The FFmpeg binary the app downloads for itself is extracted without its executable bit on macOS and Linux, so it can never be spawned.

`unzipFile` reads each entry's mode from yauzl but never applies it, so every extracted file lands as `0644`. That is harmless for publication content, but `unzipAndFindFFmpeg` returns a path to a binary that cannot be run. Everything that shells out to FFmpeg then fails with `EACCES`:

- media export
- converting a still or an audio file into a video

Windows decides executability by extension and has no bit to set, which is likely why this has gone unnoticed.

Reproduced on macOS 15 (arm64). The downloaded binary is intact — it runs correctly as soon as the bit is set by hand:

```
$ ls -la <userData>/ffmpeg/ffmpeg
-rw-r--r--  80128296  ffmpeg

$ <userData>/ffmpeg/ffmpeg -version
permission denied

$ chmod +x <userData>/ffmpeg/ffmpeg && <userData>/ffmpeg/ffmpeg -version
ffmpeg version 6.1-tessus  https://evermeet.cx/ffmpeg/
```

### Changes

- `setExecutable()` in `src-electron/main/fs.ts` — `chmod 0o755`, a no-op on Windows — exposed over IPC and preload
- `setupFFmpeg()` calls it once it has a path

The two branches of `setupFFmpeg` were restructured so both reach that call. The one that matters is the branch reusing an already-downloaded copy: it means an install left broken by an earlier version repairs itself on the next launch, rather than needing the cache cleared first.

A failed `chmod` clears the cached path, since leaving it would have the early return at the top of `setupFFmpeg` hand out the unusable binary for the rest of the session.

### Alternative considered

Applying entry modes in `unzipFile` itself would be the more general fix, but it changes permissions for every archive the app extracts — `.jwpub` content among them — for the benefit of one file that is known to need it. This keeps the blast radius to FFmpeg.

### Verification

`yarn lint`, `yarn vue-tsc --noEmit` and `yarn test:unit` all pass on this branch. One pre-existing failure in `src/utils/__tests__/api.test.ts > fetchMemorials` reproduces on an unmodified `master` and is unrelated.

### DIFF
--- a/src-electron/electron-preload.ts
+++ b/src-electron/electron-preload.ts
@@ -148,6 +148,7 @@ const electronApi: ElectronApi = {
   sendKeyTap: (k, m) => sendKeyTap(k, m),
   setAutoStartAtLogin: (v) => send('toggleOpenAtLogin', v),
   setElectronUrlVariables: (v) => send('setElectronUrlVariables', v),
+  setExecutable: (p) => invoke('setExecutable', p),
   setHardwareAcceleration: (v) => invoke('set-hardware-acceleration', v),
   setPathProbeNotificationPaths: (paths) =>
     send('setPathProbeNotificationPaths', paths),
--- a/src-electron/main/fs.ts
+++ b/src-electron/main/fs.ts
@@ -10,6 +10,7 @@ import { app, dialog } from 'electron';
 import { ensureDir, type Stats } from 'fs-extra';
 import { createReadStream, createWriteStream } from 'node:fs';
 import {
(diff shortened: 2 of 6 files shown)


sircharlo:
Thank you for this PR!

PR #8735, opened 2026-08-13
fix(scripts): make electron-rebuild script cross-platform

MatthewCoder13:
### Summary
This PR fixes the `electron-rebuild` yarn script by removing the `.cmd` extension and the explicit path to `./node_modules/.bin/`. 

### Context & Problem
Previously, the script referenced `./node_modules/.bin/electron-rebuild.cmd`, which only exists on Windows environments. This broke execution on Linux/macOS, where yarn generate binary symlinks without the `.cmd` extension.

### Solution
By invoking `electron-rebuild` directly, yarn will automatically resolve the correct executable from `node_modules/.bin` according to the host OS, making the script cross-platform.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "url": "https://github.com/sircharlo/meeting-media-manager.git"
   },
   "scripts": {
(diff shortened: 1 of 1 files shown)


sircharlo:
Thanks for the PR!

PR #9233, opened 2026-09-13
fix(crowdin): repair the exported l10n_master files, not just the source

sircharlo:
## Why

`docs:lint` on #9227 has been red since Crowdin's Italian export landed at 05:32 UTC with three drifted heading anchors in `docs/src/it/settings-guide.md` (`{#Esclusde-cbs-pubs}`, `{#interface-scorciatoie}`, `{#Media-drag-handle}`). The push-triggered auto-repair sweep fixed those three strings in Crowdin one minute later (confirmed via the API: edited in place at 05:34 UTC, still the only suggestion, still approved), but Crowdin never re-exported the file. The branch kept the corrupted export, and the 16:23 UTC master merge simply re-linted it.

This is the residual gap the workflow header already described: repairing the source can't make Crowdin refresh an already-open PR, and there's no API to force it. Until now that meant a manual worktree fix (as on #9049 and #9182).

## What

- New `repair-branch` job in `crowdin-autorepair.yml`. It runs after the source sweep on every `push` to `l10n_master` and on manual dispatch (not on the hourly schedule: the branch only changes through pushes, and each push gets its own run).
- It checks out `l10n_master`, runs `docs/utils/fix-doc-markdown.mjs` and `scripts/fix-i18n-locales.mjs` in write mode, commits as `github-actions[bot]`, and pushes with a rebase-and-retry loop for Crowdin's bursty pushes.
- The push uses the `PERSONAL_TOKEN` secret. Pushes made with `GITHUB_TOKEN` never trigger workflows, so `docs:lint` (which mergify's `translations` queue requires) would never re-run on the repaired commit. The job fails fast with a clear message if the secret is missing.
- Both jobs skip the run triggered by the repair commit itself (Crowdin authors everything else on the branch as the integration user), and that no-op run gets its own concurrency group so it can't cancel the run that made the push.
- Strings the scripts can't repair are surfaced by a `continue-on-error` step as a warning, never blocking the push of the repairs that did succeed.
- Dispatching with `apply` unchecked only reports what the branch would need.

## One-time setup

The `PERSONAL_TOKEN` repository secret dates from 2023 and no workflow uses it, so its scope and validity are unknown. Overwrite it with a new fine-grained PAT: repository access limited to this repo, permission **Contents: Read and write**, nothing else. `CROWDIN_PERSONAL_TOKEN` is the Crowdin API key and stays as is.

## Verified

- On a local worktree of the current `l10n_master`: the docs fixer changes exactly the three anchor lines, `--check` passes afterward, the i18n check is already clean, and the i18n JSON rewrite is byte-identical to Crowdin's export format for every `src/i18n/*.json`.
- The workflow parses, every embedded script passes `bash -n`, and prettier is clean.
- Not exercised end-to-end before merge: push events read the workflow from `l10n_master`, which only picks this up at the next master merge.

## After merge

To unblock #9227 right away: Actions → Crowdin auto-repair → Run workflow on `master` with `apply` checked. The new job checks out `l10n_master` explicitly, pushes the three-line fix, `docs:lint` re-runs, and mergify merges as usual.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


sircharlo:
Tested end to end before merge by dispatching this branch's workflow (the `PERSONAL_TOKEN` secret was rotated to a fine-grained PAT first):

- Dry run ([run 34778521978](https://github.com/sircharlo/meeting-media-manager/actions/runs/34778521978), `apply` unchecked): the token checked out `l10n_master`, the docs fixer reported the three Italian anchors, the i18n fixer reported nothing, and the report/push steps were skipped as intended.
- Real run ([run 34778719437](https://github.com/sircharlo/meeting-media-manager/actions/runs/34778719437)): pushed 313b0a5 to `l10n_master` (`docs/src/it/settings-guide.md`, 3 lines). The PAT push triggered the PR checks as designed: `docs:lint` on #9227 is green on the new head, and the push-triggered run of the old single-job workflow on `l10n_master` ran its sweep as before.

So #9227 is unblocked already; merging this PR makes the branch repair automatic for future exports (push events pick up the new job once master is merged into `l10n_master`).

